PR #13, opened 2017-08-24
Add control panel chart

andyhd:


charts/control-panel/templates/deployment.yml
```diff
@@ -0,0 +1,45 @@
...
+          ports:
+            - name: http
+              containerPort: 8000
+          env:
+            - name: ALLOWED_HOSTS
+              value: "*"
+              #valueFrom:
+                #fieldRef:
+                  #fieldPath: status.podIP
+            - name: DB_HOST
```
kerin: All DB connection params should be handled as Kubernetes secrets, and have the per-env values stored in git-crypted chart env config files.

charts/control-panel/templates/service.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ template "fullname" . }}
+  labels:
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}"
+    app: {{ template "fullname" . }}
+spec:
+  sessionAffinity: ClientIP
```
kerin: We shouldn't need this - this defines sticky sessions based on the client's IP, but as this is stateless we should drop it.

charts/control-panel/templates/deployment.yml
```diff
@@ -0,0 +1,45 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: {{ template "fullname" . }}
+  labels:
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    app: {{ template "fullname" .}}
+spec:
+  replicas: 1
```
kerin: Should probably run everything with 3 replicas (one per AZ) by default

charts/control-panel/templates/ingress.yml
```diff
@@ -0,0 +1,18 @@
...
+metadata:
+  name: {{ template "fullname" . }}
+  labels:
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    app: {{ template "fullname" . }}
+  annotations:
+    kubernetes.io/ingress.class: "nginx"
+spec:
+  rules:
+  - host: "control-panel.services.{{ .Values.ToolsDomain }}"
```
kerin: This seems misnamed? Shouldn't it be:
```
  - host: "control-panel.{{ .Values.ServicesDomain }}"
```
?

charts/control-panel/values.yaml
```diff
@@ -0,0 +1,9 @@
+ToolsDomain: dev.mojanalytics.xyz
```
kerin: This should probably be an empty string here, and `ServicesDomain: services.dev.mojanalytics.xyz` in the per-env chart config

charts/control-panel/values.yaml
```diff
@@ -0,0 +1,9 @@
...
+API:
+  Image:
+    Repository: quay.io/mojanalytics/control-panel
+    Tag: 0.1.0
+    PullPolicy: Always
```
kerin: We should try to avoid using `Always` in favour of `IfNotPresent`, which requires that a new version tag is always used when a build happens

charts/control-panel/values.yaml
```diff
@@ -0,0 +1,9 @@
...
+  Environment:
+    DB_HOST: dev-control-panel-db.cyfs8aeszqyh.eu-west-1.rds.amazonaws.com
```
kerin: Should be an empty string here and stored in the chart env config instead

charts/control-panel/templates/deployment.yml
andyhd: See https://github.com/ministryofjustice/analytics-platform-config/pull/6

PR #16, opened 2017-09-11
cpanel: Control Panel API can assume an IAM role

xoen:
### What/Why?
This is so that the Control Panel API instances can have permissions to
perform certain operations (e.g. create S3 bucket, etc...)

### Ticket
https://trello.com/c/oNMQLEfF/360-add-iam-role-support-to-control-panel-api-helm-chart

### Related to
 - Role added: https://github.com/ministryofjustice/analytics-platform-ops/pull/106
 - cpanel helm chart config: https://github.com/ministryofjustice/analytics-platform-config/pull/8


xoen:
I also added some environment variables the control panel API will need once James work on the AWS stuff is done.

PR #32, opened 2017-10-09
cpanel: Added `SENTRY_DSN` environment variable value

xoen:
### What

This is to pass the Sentry credentials to the control-panel-api: https://github.com/ministryofjustice/analytics-platform-control-panel/pull/34

### Part of ticket

https://trello.com/c/G8qny9WN/400-cp-api-set-up-sentry-for-exceptions

### DIFF
--- a/charts/cpanel/templates/deployment.yaml
+++ b/charts/cpanel/templates/deployment.yaml
@@ -60,6 +60,8 @@ spec:
               value: "{{ .Values.API.Environment.K8S_WORKER_ROLE_NAME }}"
             - name: LOGS_BUCKET_NAME
               value: "{{ .Values.API.Environment.LOGS_BUCKET_NAME }}"
+            - name: SENTRY_DSN
+              value: "{{ .Values.API.Environment.SENTRY_DSN }}"
           readinessProbe:
             httpGet:
               path: /auth/login
--- a/charts/cpanel/templates/deployment.yaml
+++ b/charts/cpanel/templates/deployment.yaml
@@ -61,7 +61,10 @@ spec:
             - name: LOGS_BUCKET_NAME
               value: "{{ .Values.API.Environment.LOGS_BUCKET_NAME }}"
             - name: SENTRY_DSN
-              value: "{{ .Values.API.Environment.SENTRY_DSN }}"
+              valueFrom:
+                secretKeyRef:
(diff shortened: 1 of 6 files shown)


xoen:
A small one for you @andyhd :)

charts/cpanel/templates/deployment.yaml
```diff
@@ -60,6 +60,8 @@ spec:
               value: "{{ .Values.API.Environment.K8S_WORKER_ROLE_NAME }}"
             - name: LOGS_BUCKET_NAME
               value: "{{ .Values.API.Environment.LOGS_BUCKET_NAME }}"
+            - name: SENTRY_DSN
+              value: "{{ .Values.API.Environment.SENTRY_DSN }}"
```
andyhd: Should this be a secret?

xoen: Mmmm, let me think about it. What do you mean?

It needs to come from the values and ends in the environment variable. Secrets can be used to store passwords, etc... (despite not being encrypted really) while waiting to be used by different pods, etc...but in this case the end result would still be to end up in the pod environment variable.

xoen: What I mean is that I could store the value coming from the value in a secret and then setting the variable from the secret but what does this level of indirection gives me here?

@kerin - is there any benefit also storing this in a secret in this case?

xoen: @andyhd FYI: Had a quick chat with Kerin to ask some questions regarding secrets in helm/k8s, etc...You made a good point. I'll amend this chart so this will come from a secret 👍

kerin: @xoen @andyhd FYI encryption of secrets is an alpha feature in k8s 1.7 and beta in 1.8, so this has value once we've upgraded our clusters: https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/

xoen:
@andyhd GitHub "lost" the previous conversation on the secret as I changed that part of the code now to move `SENTRY_DSN` into a secret as you suggested: https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/32/commits/1986d26594de7378194876f82078f8fa21cbf241

xoen:
Cheers @andyhd :)

PR #33, opened 2017-10-13
Add chart for NFS user home dir backup cronjob

kerin:
## What

Adds cronjob and configmap to run `s3cmd sync` every hour to backup NFS home dirs to S3

File patterns excluded from backups are listed in `charts/nfs-backup/templates/configmap.yml`

## How to review
1. Checkout this branch and associated PRs in [analytics-platform-ops](https://github.com/ministryofjustice/analytics-platform-ops/pull/112) and [analytics-platform-config](https://github.com/ministryofjustice/analytics-platform-config/pull/14)
2. Apply Terraform changes from `analytics-platform-ops` repo to ensure the S3 bucket and IAM resources exist
3. `helm install nfs-backup -f ../analytics-platform-config/chart-env-config/ENV/nfs-backup.yml`
4. Confirm that the cronjob is installed with `kubectl get cronjob`
5. Confirm that the job successfully runs on the hour by checking logs with `kubectl logs $POD_NAME`


charts/nfs-backup/templates/configmap.yml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ template "fullname" . }}
+  labels:
+    chart: {{ .Chart.Name }}-{{ .Chart.Version }}
+data:
+  exclude_patterns: |
```
RobinL: I think there's a pretty solid argument to say we shouldn't be backing up any data files that are larger than `x`.  I think `x`  might be something like 100Mb.   

Relevant file extensions would be .Rdata, .xlsx, and .csv at minimum.

If we don't like that, we should almost certainly not backup .Rdata files at all.  This is R's 'memory dump' that allows you to restore everything in memory.

Or if we do back this stuff up, retention should be only for short periods.

Justification:  
Users' home directory is absolutely not the place for them to be storing large datasets with sensitive data.  We probably also don't want this all being dumped into a single s3 bucket, that would then potentially contain loads of sensitive data from across the justice system.

charts/nfs-backup/templates/configmap.yml
```diff
@@ -0,0 +1,17 @@
...
+    \/\.rstudio\/
+    \/R\/library\/
+    \/packrat\/
+    \/.git\/
+    \/.cache\/
+    \/.Rproj.user\/
```
RobinL: Also might want want to add `\/.Rhistory\/`, which stores a list of the commands the user has sent to the console.

charts/nfs-backup/templates/configmap.yml
kerin: @RobinL see updated exclusion list

charts/nfs-backup/templates/configmap.yml
```diff
@@ -0,0 +1,20 @@
...
+  exclude_patterns: |
+    \/\.rstudio\/
+    \/R\/library\/
+    \/packrat\/
+    \/\.git\/
+    \/\.cache\/
+    \/\.Rproj\.user\/
+    \/\.RData$
+    \/\.Rhistory$
+    \/id_rsa$
```
xoen: Just started to review this, so sorry for the stupid questions.

1) I imagine these files to exclude from the backups, right?
2) Do we want to ignore `id_rsa` files? I don't know much about the others, I suppose the question is, if something goes wrong and we re-create the users' home from these backups, will the user miss any of these files/notice they're missing?

charts/nfs-backup/templates/cronjob.yml
```diff
@@ -0,0 +1,51 @@
...
+                mountPath: "/homes"
+              - name: exclude-patterns
+                mountPath: /tmp/s3cmd
+          restartPolicy: OnFailure
+          volumes:
+          - name: nfs-homes
+            persistentVolumeClaim:
+              claimName: nfs-homes
+          - name: exclude-patterns
+            configMap:
```
xoen: You know I'm not a k8s expert, don't we need the `key: exclude_patterns`? Is it going to create a volume with these two files and the values as content maybe?

charts/nfs-backup/templates/cronjob.yml
```diff
@@ -0,0 +1,51 @@
...
+          name: nfs-backup
+          annotations:
+            iam.amazonaws.com/role: {{ .Values.AWS.IAMRole }}
+        spec:
+          containers:
+          - image: "{{ .Values.Image.Repository }}:{{ .Values.Image.Tag }}"
+            imagePullPolicy: {{ .Values.Image.PullPolicy }}
+            name: s3cmd
+            args:
+              - sync
```
xoen: I'm starting to have a look at s3cmd, etc...how would the sync work when files will be deleted from the destination S3 bucket? Will the backup re-upload them?

charts/nfs-backup/templates/cronjob.yml
kerin: If a key isn't specified then all key/value pairs in the configmap will be projected into files at the specified path, with the key name as the file name - so here `/tmp/s3cmd/exclude_patterns` will be created, as referenced in the s3cmd `args` list.

charts/nfs-backup/templates/cronjob.yml
kerin: Yes - if files expire from the s3 bucket because they've not been updated during the retention period then they'll get re-uploaded on the next backup run. This is what we want though I think - it means that all files are retained for their lifetime in NFS + 30 days.

charts/nfs-backup/values.yaml
```diff
@@ -0,0 +1,9 @@
+AWS:
+  IAMRole: ""
+  S3:
+    BucketName: ""
+Image:
+  Repository: samepagelabs/s3cmd
+  Tag: "1.6.1"
+  PullPolicy: IfNotPresent
+Schedule: "0 * * * *"
```
xoen: This means every hour right? (interesting that the schedule of the `dev` pod `nfs-backup-nfs-backup` is `20 * * * *`)

Seems like k8s supports Cron format strings like `@hourly`, I _personally_ prefer them as they read better and avoid confusion (but you don't have to change it!)

kerin: Oh the schedule in dev is at 20 past the hour because I was testing a change at 11.19.

Didn't realise that kubernetes supported meta format strings - that's much more readable, I'll change it.

PR #43, opened 2017-10-25
Add SAML_PROVIDER env var to cpanel

andyhd:


charts/cpanel/Chart.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: Analytics Platform Control Panel API webapp
 name: cpanel
-version: 0.5.0
+version: 0.5.1
```
xoen: I don't mind if we only bump the patch here. I wonder if this is the case of bumping the minor (according to [semantic versioning](http://semver.org)).

I suppose depends on if we consider this a bug fix or not - for some reason adding the ability to pass something new to the app seems a functionality to me - but as I said above, don't mind.

charts/cpanel/templates/deployment.yaml
```diff
@@ -60,6 +60,8 @@ spec:
               value: "{{ .Values.API.Environment.K8S_WORKER_ROLE_NAME }}"
             - name: LOGS_BUCKET_NAME
               value: "{{ .Values.API.Environment.LOGS_BUCKET_NAME }}"
+            - name: SAML_PROVIDER
```
xoen: Maybe would be good to document this param in the README and maybe add a line to the CHANGELOG as well.

I see we concatenate this with `IAM_ARN_BASE` in the code, does it mean this is the provider role name now? If that's the case at cost of having slightly longer param name I'd probably prefer to call it something more explicit as `SAML_PROVIDER_ROLE_NAME`, what do you reckon?

charts/cpanel/Chart.yaml
andyhd: It's a non-breaking bug fix, so I think it meets the definition of a semver PATCH: "PATCH version when you make backwards-compatible bug fixes."

charts/cpanel/templates/deployment.yaml
andyhd: Yep to both

xoen: Actually, this is not a role name but a provider name so maybe call it `SAML_PROVIDER_NAME` doesn't give you much 👍

andyhd: In that case, I will just leave it 😴

PR #44, opened 2017-10-26
cpanel 0.6.0: Added OIDC_* settings

xoen:
These are required by the Control Panel API to make requests
to Auth0 and to verify validity of JWT tokens.


charts/cpanel/Chart.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: Analytics Platform Control Panel API webapp
 name: cpanel
-version: 0.5.1
+version: 0.6.0
```
andyhd: I'd say this is only a patch release, because it is backwards compatible - you're only adding env vars and not removing any.

PR #50, opened 2017-10-31
Adds .gitignore, pre-push hook and pre-commit hook setup to config-user chart

kerin:
**NOTE**

This PR is not yet 100% complete due to the need for a new Dockerfile/built container, to remove the need to perform an `apt-get install` during git config.

To review/test:

**In dev cluster:**

Delete your existing `config-user` release:
`$ helm delete --purge config-user-$USERNAME`

Reinstall chart from this repo:

`$ helm install charts/config-user --name config-user-$USERNAME --set Username=$USERNAME --namespace user-$USERNAME`

Watch your running pods until the `config-user` pod completes:

`$ kubectl get pods -n user-$USERNAME -w`

Log into your RStudio instance, check out a git repository (it must be a new repo as config changes aren't retroactive), and confirm that you're unable to:

- check in `.csv`, `.xls`, etc. files 
- push to a repo not owned by `moj-analytical-services` or `ministryofjustice`
- commit a file larger than 5MB


charts/config-user/files/pre-commit
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh
+#
+# An example hook script to verify what is about to be committed.
+# Called by "git commit" with no arguments.  The hook should
+# exit with non-zero status after issuing an appropriate message if
+# it wants to stop the commit.
+#
+# To enable this hook, rename this file to "pre-commit".
```
andyhd: This comment is no longer relevant to the code

charts/config-user/files/pre-push
```diff
@@ -0,0 +1,48 @@
...
+#
+# If pushing without using a named remote those arguments will be equal.
+#
+# Information about the commits which are being pushed is supplied as lines to
+# the standard input in the form:
+#
+#   <local ref> <local sha1> <remote ref> <remote sha1>
+#
+# This sample shows how to prevent push of commits where the log message starts
+# with "WIP" (work in progress).
```
andyhd: This comment is no longer relevant

charts/config-user/files/pre-push
```diff
@@ -0,0 +1,48 @@
...
+
+echo "
+You are attempting to push a repository to an unknown remote user or GitHub
+organisation \"${org}\". Only repositories owned by the following GitHub
+organisations are permitted as push targets:
+
+$(printf '%s\n' "${allowed_orgs[@]}")
+
+Please update your .git/config file and modify your remote.
+"
```
andyhd: Does this work? I though multi-line strings in bash were done with heredocs, eg:
```sh
echo <<"EOF"
...
EOF;
```

charts/config-user/files/git-config.sh
```diff
@@ -0,0 +1,57 @@
...
+if [ $# -lt 3 ]; then
+  echo 1>&2 "usage: $0 USERNAME FULLNAME EMAIL"
+  exit 2
+fi
+
+USERNAME=$1
+FULLNAME=$2
+EMAIL=$3
+
+SCRIPT_PATH="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
andyhd: Will this change the directory the script was run from, or does it only affect the context of the `$()`? Would it be better to use something like `pushd` and `popd`?

charts/config-user/files/git-config.sh
```diff
@@ -0,0 +1,57 @@
...
+
+USER_HOME=/home/${USERNAME}
```
andyhd: Do these need to be escaped/quoted/sanitized? Can a username contain `../`? (probably not)

charts/config-user/files/git-config.sh
```diff
@@ -0,0 +1,57 @@
...
+if [ ! -f $GIT_IGNORE ]; then
+  cp $SCRIPT_PATH/gitignore $GIT_IGNORE
+fi
+
+
+# ~/.git-templates
+GIT_TEMPLATES=$USER_HOME/.git-templates
+
+if [ ! -d $GIT_TEMPLATES ]; then
+  mkdir -p $GIT_TEMPLATES
```
andyhd: Doesn't `mkdir -p` create directories (and their parents) if they don't exist? I would have thought the `if [ ! -d` redundant.

charts/config-user/files/pre-commit
```diff
@@ -0,0 +1,29 @@
...
+        HAS_ERROR="1"
+    fi
+done
+
+if [ "$HAS_ERROR" != "" ]; then
+    echo "Can't commit due to oversize files, please unstage before committing." >&2
+    exit 1
+fi
+
+exit 0
```
xoen: Do you need this? I always thought it was implicitly returning `0` when nothing went wrong.

charts/config-user/files/git-config.sh
kerin: Yeah that cd's into the directory containing the script, then sets that path as a variable so the rest of the script can refer to it. It's probably not the safest way to get the directory of a script, but in this case I think it's OK as there are no relative paths used.

charts/config-user/files/git-config.sh
kerin: Usernames are github usernames, and so can't contain anything other than `a-zA-Z0-9-_`. That said it'd probably be safer to check that the path exists and begins with `/home/` before proceeding.

charts/config-user/files/git-config.sh
kerin: `-p` will create intermediary directories if they don't already exist, but it'll still error out if the path exists, hence the check.

charts/config-user/files/pre-push
```diff
@@ -0,0 +1,48 @@
...
+)
+
+for i in "${allowed_orgs[@]}"; do
+    if [[ "$i" == "$org" ]]; then
+        exit 0;
+    fi
+done
+
+echo "
+You are attempting to push a repository to an unknown remote user or GitHub
```
xoen: Unless I'm missing something, we're not checking the user in this script, right? The message says otherwise.

Also I imagine this would prevent pushing to a user branch (e.g. [https://github.com/kerin/something]()) - it may be what we want.

charts/config-user/files/pre-commit
kerin: You're no longer relevant to the code.

charts/config-user/files/pre-push
kerin: Yeah multiline strings in double quotes work - maybe heredoc would be cleaner though.

charts/config-user/files/pre-push
kerin: Yep, this is to ensure they only push to MOJ-owned repositories so that public/private status can be controlled at the admin level - we're specifically preventing users from pushing to their own github accounts.

charts/config-user/templates/configmap-git-config.yaml
```diff
@@ -1,23 +1,8 @@
...
-    if [ -f $GIT_CONFIG ]; then
-      exit 0
-    fi
-
-    echo -e "[user]" > $GIT_CONFIG
-    echo -e "\tname = $FULLNAME" >> $GIT_CONFIG
-    echo -e "\temail = $EMAIL" >> $GIT_CONFIG
-
-    chown 1001:staff $GIT_CONFIG
+{{ (.Files.Glob "files/*").AsConfig | indent 2 }}
```
xoen: Interesting one :)

charts/config-user/files/gitignore
```diff
@@ -0,0 +1,79 @@
...
+# it's better to unpack these files and commit the raw source
+# git has its own built in compression methods
+*.7z
+*.dmg
+*.gz
+*.iso
+*.jar
+*.rar
+*.tar
+*.zip
```
xoen: Worth pointing out (although this is copied from somewhere else), as Andy said earlier, that this is still not catching things like `.ZIP`, etc..which could potentially be used by the user to push sensitive data...but they can always do that if they really want to.

charts/config-user/files/git-config.sh
kerin: ...although the `cd` happens in a subshell, so doesn't change the current directory in the script itself

charts/config-user/files/pre-push
xoen: I see. In that case I think we should drop the bit of the message talking of unknown users to avoid possible confusion.

PR #68, opened 2018-01-02
Role-binding to make user's admins in their own namespaces

djsd123:
This assigns the admin user-facing role to a user, scoped to the user's namespace.


xoen:
Actually, I would move the `RoleBinding` resource in its own file (`role-binding.yml`) for consistency with the rest of the helm charts, and also to make it easy to find.

djsd123:
@xoen Thanks.  I was thinking maybe `rbac.yml` also just in case we wanted to extend later.  Cross that bridge .......

PR #73, opened 2018-01-12
OIDC Username Claim

djsd123:
The Kubernetes API prefixes the 'IssuerURL' to all UsernameClaims other than email.
To use the nickname claim, we must prefix the subject's name with the 'IssuerURL' until which point kops supports the --oidc-username-prefix flag.
See: https://kubernetes.io/docs/admin/authentication/#configuring-the-api-server

Also, see Aldo's PR where he added support for --oidc-username-prefix
to kops: kubernetes/kops#4085

### DIFF
--- a/charts/init-user/templates/role-binding.yml
+++ b/charts/init-user/templates/role-binding.yml
@@ -10,4 +10,4 @@ roleRef:
 subjects:
   - apiGroup: rbac.authorization.k8s.io
     kind: User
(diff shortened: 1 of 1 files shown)


xoen:
As discussed yesterday afternoon, let's keep this around as proof of concept to be able to test in the `dev` cluster. Hopefully we'll be able to override the username prefix soon-ish using KOPS.

PR #81, opened 2018-01-24
kube2iam v1.0.0: Added Service account, to make it work when RBAC enabled

xoen:
k8s resources copied from https://github.com/jtblin/kube2iam/pull/99

### DIFF
--- a/charts/README.md
+++ b/charts/README.md
@@ -47,6 +47,17 @@ $ helm install charts/kube-dashboard -f chart-env-config/dev/kube-dashboard.yml 
 ```
 
 
+## kube2iam
+
+Responsible for attaching IAM roles to pods with the `iam.amazonaws.com/role`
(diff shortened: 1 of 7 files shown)


xoen:
Cheers @kerin - I may try to install this in `dev` in a bit or tomorrow and see if it fixes that errors getting the AWS creds 👍

xoen:
For the record. kube2iam's `ServiceAccount` lives in the `default` namespace and the corresponding namespace in the `ClusterRoleBinding`'s subject/service account needs to be `default` as well for this to work.

xoen:
FYI: Merging this asI  Tested in `dev` and I can now get the AWS credentials from within the CP-API pod.

**NOTE**: I installed this new version of the `kube2iam` chart in `dev` but **NOT** in `alpha`.

PR #88, opened 2018-02-07
Add idler chart

andyhd:
* Run idler script as a cron job (default every night at 10pm)

### DIFF
--- a/charts/idler/values.yaml
+++ b/charts/idler/values.yaml
@@ -1,0 +1,13 @@
+# Docker image version
+image: quay.io/mojanalytics/idler:latest
+
+# Schedule when to run
+#   min    hour   day    month (Sun-Sat)
+# "(0-59) (0-23) (1-31) (1-12) (0-6)"
+# See https://kubernetes.io/docs/user-guide/cron-jobs/#schedule
+schedule: "0 22 * * *"
+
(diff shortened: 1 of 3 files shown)


djsd123:
It looks good.  Have you deployed it to an `RBAC` enabled cluster yet?

andyhd:
Not yet

charts/idler/values.yaml
```diff
@@ -0,0 +1,13 @@
...
+# Schedule when to run
+#   min    hour   day    month (Sun-Sat)
+# "(0-59) (0-23) (1-31) (1-12) (0-6)"
+# See https://kubernetes.io/docs/user-guide/cron-jobs/#schedule
+schedule: "0 22 * * *"
+
+# Concurrency Policy
+# Allow|Forbid|Replace
+# See https://kubernetes.io/docs/user-guide/cron-jobs/#concurrency_policy
+concurrencyPolicy: Forbid
```
xoen: I wonder if we need a value for this. I guess it doesn't harm to have configurability but you know, YAGNI.

(from what I can tell by looking at the idler script it should cause mayhem if more than one runs concurrently but I agree with you it's safer to have only one job of this time running just to be on the safe side 👍 )

PR #90, opened 2018-02-07
Add RBAC role, binding and service account for idler

andyhd:


charts/idler/templates/clusterrole.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ .Release.Name }}
+  labels:
+    app: {{ .Release.Name }}
+rules:
+  - apiGroups:
+      - ""
```
djsd123: Empty String means `v1`.  Sorry, I should have been explicit in my PR.  I think you want `apps` or `extensions`

PR #91, opened 2018-02-08
Kubernetes MetaData Plugin

djsd123:
See: https://github.com/kubernetes/kubernetes/issues/53022#issuecomment-332150416

Also

See: https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter

Symlink `/var/log/containers` has been deprecated in favour of `/var/log/pods`
for our version of the docker daemon for kubernetes metadata.

No noticeable symptoms other than log errors.

### DIFF
--- a/charts/fluentd/values.yaml
+++ b/charts/fluentd/values.yaml
@@ -146,7 +146,7 @@ fluentdconffiles:
       </exclude>
     </filter>
 
-    <filter kubernetes.**>
+    <filter kubernetes.var.log.pods.**.log>
       @type kubernetes_metadata
     </filter>
   fluent.conf: |


charts/fluentd/values.yaml
```diff
@@ -146,7 +146,7 @@ fluentdconffiles:
       </exclude>
     </filter>
 
-    <filter kubernetes.**>
+    <filter kubernetes.var.log.pods.**.log>
```
xoen: Sorry, not much familiar with fluentd, who handle this `kubernetes_metadata` type?

xoen: So a Docker change moved where pod logs are stored right? And with this dot (`.`) notations we fix the submission of pods' logs, right?

djsd123: Yes, that's how we specify it according to the [docs](https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter).  I had to double check too.  I'm just being explicit here with the path and fluentd no longer complains about it.  See: https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/issues/105

Those guys know way more than i do

PR #96, opened 2018-02-14
RStudio 1.3.0: Pass env variable necessary for hosted login page logout

xoen:
### What
Set environment variables used by [rstudio-auth-proxy](https://github.com/ministryofjustice/analytics-platform-rstudio-auth-proxy/pull/12/files):
- `APP_PROTOCOL` will always be `https`
- `APP_HOST` is something like `joe-rstudio.tools.example.com`

### NOTES
This needs to be merged before the [rstudio-auth-proxy changes](https://github.com/ministryofjustice/analytics-platform-rstudio-auth-proxy/pull/12/files) (if not logout would be broken, rest should still be fine so not a massive deal really)

### Part of ticket
https://trello.com/c/t90fEmBs/647-4-update-rstudio-auth-proxy-to-use-hosted-auth0-login

### DIFF
--- a/charts/rstudio/CHANGELOG.md
+++ b/charts/rstudio/CHANGELOG.md
@@ -1,0 +1,12 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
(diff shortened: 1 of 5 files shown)


xoen:
Cheers Andy

PR #99, opened 2018-02-14
Jupyter-Lab: Initial Commit

djsd123:
Jupyter Lab Helm Chart

Deploys Jupyter lab data-science notebook [stack](https://github.com/jupyter/docker-stacks/tree/master/datascience-notebook) 

Shares the same storage(PVC) as rstudio-server.

Uses image: https://github.com/ministryofjustice/analytics-platform-jupyter-notebook

### DIFF
--- a/charts/jupyter-lab/README.md
+++ b/charts/jupyter-lab/README.md
@@ -1,0 +1,33 @@
+# Jupyter-Lab Helm Chart
+
+
+## Installing the Chart
+
+To install an jupyter instance for the user specified in the Username variable (Github username):
+
+```bash
+$ helm install charts/jupyter-lab -f chart-env-config/ENV/jupyter.yml --set Username=USERNAME --namespace user-USERNAME --name=USERNAME-jupyter
+```
+
+The instance will be available in <https://USERNAME-jupyter.tools.ENV.mojanalytics.xyz>.
+
+**NOTE**: Change the environment config file to deploy in a different environment
+          (the URL will change accordingly)
+
+
+## Upgrading the Chart
+
+To upgrade a user jupyter chart:
+```bash
+$ helm upgrade USERNAME-jupyter charts/jupyter-lab -f chart-env-config/ENV/jupyter.yml --set Username=USERNAME
+```
+
+
+## Configuration
+
+Listing only the required params here. See `/chart-env-config/` for more
+details.
+
+| Parameter  | Description     | Default |
+| ---------- | --------------- | ------- |
+| `Username` | Github username | ``      |
--- a/charts/jupyter-lab/templates/deployment.yaml
+++ b/charts/jupyter-lab/templates/deployment.yaml
@@ -1,0 +1,97 @@
+apiVersion: apps/v1beta2
+kind: Deployment
+metadata:
+  name: {{ template "fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    chart: {{ .Chart.Name }}-{{ .Chart.Version }}
+    app: {{ .Chart.Name }}
+spec:
+  selector:
+    matchLabels:
+      app: {{ .Chart.Name }}
+  replicas: 1
+  strategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxSurge: 1
+      maxUnavailable: 0
+  template:
+    metadata:
+      labels:
+        app: {{ .Chart.Name }}
+      annotations:
+        iam.amazonaws.com/role: {{ .Values.aws.iamRole }}
+    spec:
+      serviceAccountName: {{ .Values.Username }}-jupyter
+      containers:
+        - name: {{ .Chart.Name }}-auth-proxy
+          image: {{ .Values.authProxy.image }}:{{ .Values.authProxy.tag }}
+          imagePullPolicy: {{ .Values.authProxy.imagePullPolicy }}
+          ports:
+            - name: http
+              containerPort: {{ .Values.authProxy.containerPort }}
+          env:
+            - name: TARGET_URL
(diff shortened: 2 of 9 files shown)


charts/jupyter-lab/templates/deployment.yaml
```diff
@@ -0,0 +1,97 @@
...
+      maxSurge: 1
+      maxUnavailable: 0
+  template:
+    metadata:
+      labels:
+        app: {{ .Chart.Name }}
+      annotations:
+        iam.amazonaws.com/role: {{ .Values.aws.iamRole }}
+    spec:
+      serviceAccountName: {{ .Values.Username }}-jupyter-lab
```
xoen: Minor point, it may be more predictable if the service account name matches the subdomain, e.g. `USER-jupyter`, not a big deal obviously :)

charts/jupyter-lab/templates/deployment.yaml
```diff
@@ -0,0 +1,97 @@
...
+                secretKeyRef:
+                  name: {{ template "fullname" . }}
+                  key: auth0_client_secret
+            - name: COOKIE_SECRET
+              valueFrom:
+                secretKeyRef:
+                  name: {{ template "fullname" . }}
+                  key: cookie_secret
+            - name: NICKNAME_RE
+              value: {{ .Values.Username }}
```
xoen: Don't remember how the `NICKNAME_RE` works in the auth-proxy. This would only proxy for an exact match between the username from Auth0 and `.Values.Username`, right?

charts/jupyter-lab/templates/deployment.yaml
```diff
@@ -0,0 +1,97 @@
...
+              port: http
+            initialDelaySeconds: 5
+            periodSeconds: 5
+          volumeMounts:
+            - name: workspace
+              mountPath: /home/jovyan
+          resources:
+{{ toYaml .Values.jupyter.resources | indent 12 }}
+      volumes:
+        - name: workspace
```
xoen: Just a matter of taste here I suppose, wouldn't be more revealing if this volume was named `user-home` or something like that? Do you reckon is a silly idea?

charts/jupyter-lab/values.yaml
```diff
@@ -0,0 +1,38 @@
...
+aws:
+  iamRole: ""
+  defaultRegion: "eu-west-1"
+
+service:
+  port: 80
+
+authProxy:
+  image: quay.io/mojanalytics/auth-proxy
+  tag: v0.0.2
```
xoen: I've noticed this is not the latest tag (semantically speaking) [in Quays](https://quay.io/repository/mojanalytics/auth-proxy?tab=tags) - any reason why we're not using the latest (`v0.2.0`)?

Even weirdier, I notice that `v0.2.0` is 7 months old but `v0.1.0` is 9 days old - Don't know what happened there. In theory `v0.2.0` > `v0.1.0` therefore, `v0.2.0` should be more recent no? What am I missing?

charts/jupyter-lab/values.yaml
```diff
@@ -0,0 +1,38 @@
...
+      cpu: 100m
+      memory: 128Mi
+    requests:
+      cpu: 25m
+      memory: 64Mi
+  auth0_domain: ""
+  auth0_client_id: ""
+
+jupyter:
+  image: quay.io/mojanalytics/datascience-notebook
```
xoen: I was surprised to not find the `mojanalytics/jupyter` image in Quay, any reason to call this `datascience-notebook` rather than simply `jupyter`?

charts/jupyter-lab/README.md
```diff
@@ -0,0 +1,33 @@
+# Jupyter-Lab Helm Chart
```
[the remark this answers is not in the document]
djsd123: Thanks

charts/jupyter-lab/templates/deployment.yaml
djsd123: Sure, if you think it's more clear, I can change it.

charts/jupyter-lab/templates/deployment.yaml
djsd123: The way I am using it, yes.  It's a regex but we're just using it for an exact match.  I think.

charts/jupyter-lab/templates/deployment.yaml
djsd123: Sure.

charts/jupyter-lab/values.yaml
djsd123: You'll have to speak to @andyhd about the versioning.

As for not using the latest tag? https://github.com/ministryofjustice/analytics-platform-auth-proxy/issues/4

I intend to look into it and hopefully fix with Andy's and your's help.

charts/jupyter-lab/values.yaml
djsd123: sure, I'll re-tag it.

charts/jupyter-lab/values.yaml
andyhd: The latest version of the auth proxy appears to have a bug which prevents jupyter lab from working. There shouldn't be any change to functionality in between versions, so it's safe to continue using the old version while we debug the new one.

xoen: Oh I see, I now remember you mentioning it may have been caused by the refactoring last week.

I'm currently in `rstudio-auth-proxy` land but at some point I'll update the generic `auth-proxy` image as well to use Auth0 hosted login page: Do you remember exactly what is not working/(roughly) what's the bug?

(BTW, happy for this to use the old/working `auth-proxy` for the time being 👍 )

PR #100, opened 2018-02-14
Create User Home Namespace

djsd123:
The persistent volume claim `nfs-homes` lives in the default namspace.  Running the
Job works fine if the namespace in your context is set to default but in
the event that it is not.  The job will fail.


jmoz:
LGTM 👍

djsd123:
Ta @jmoz, @xoen

PR #104, opened 2018-02-16
RStudio v1.3.0 proxy to use Auth0 hosted login page

xoen:
See quay.io/mojanalytics/rstudio-auth-proxy tag v1.3.0


xoen:
Thanks Andy

PR #105, opened 2018-02-16
RBAC: CPANEL

djsd123:
In the backgroud the control panel invokes `helm` commands to manage a
user's applications like rstudio. `helm` itself requires that the
account invoking it be able to discover `tiller`. To discover `tiller` the account
at the very least will need to be able to __list__ `pods` in `kube-system` (perhaps
because helm assumes that is where tiller is?) and forward ports.


djsd123:
Thanks @andyhd

xoen:
Thanks @djsd123 - Good one 👍

djsd123:
Thank you @xoen 🔢

PR #107, opened 2018-02-20
init-user: Bumped version

xoen:
The version didn't change which means the repository index rebuild
didn't get triggered.


xoen:
Approved by @djsd123 offline

PR #108, opened 2018-02-21
Add unidler chart

andyhd:


charts/unidler/templates/clusterrolebinding.yaml
```diff
@@ -0,0 +1,15 @@
...
+    app: {{ .Release.Name }}
+subjects:
+  - kind: ServiceAccount
+    name: {{ .Release.Name }}
+    namespace: default
+    apiGroup: ""
+roleRef:
+  kind: ClusterRole
+  name: {{ .Release.Name }}
+  apiGroup: ""
```
djsd123: Should this be `rbac.authorization.k8s.io`

charts/unidler/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Release.Name }}
```
djsd123: Think you should probably add the namespace you want this account to be created in.  Going by the ClusterRolebinding it should be `default` if that was your intention.

charts/unidler/templates/deployment.yaml
```diff
@@ -0,0 +1,25 @@
...
+      labels:
+        app: {{ .Release.Name }}
+    spec:
+      serviceAccountName: {{ .Release.Name }}
+      containers:
+      - name: {{ .Chart.Name }}
+        image: "{{ .Values.image.name }}:{{ .Values.image.tag }}"
+        imagePullPolicy: {{ .Values.image.pullPolicy }}
+        ports:
+        - name: {{ .Values.service.name }}
```
djsd123: Is this Value missing?

charts/unidler/templates/deployment.yaml
```diff
@@ -0,0 +1,25 @@
...
+      containers:
+      - name: {{ .Chart.Name }}
+        image: "{{ .Values.image.name }}:{{ .Values.image.tag }}"
+        imagePullPolicy: {{ .Values.image.pullPolicy }}
+        ports:
+        - name: http
+          containerPort: {{ .Values.service.internalPort }}
+        env:
+          - name: LOG_LEVEL
+            value: "INFO"
```
xoen: (Sorry for the late comments, I know this is already merged)

Not a problem, just a thought, we may these kind of variables in the `values.yaml` so by default would be `INFO` but we could potentially easily override if needed.

As I said, just a general thought, nothing we really need to change.

charts/unidler/values.yaml
```diff
@@ -0,0 +1,14 @@
+# Docker image
+image:
+  name: quay.io/mojanalytics/unidler
+  tag: v0.1.0
```
xoen: I see that this tag is [not present in Quay](https://quay.io/repository/mojanalytics/unidler?tab=tags), I know you're still playing/testing the unidler so maybe you're already aware of it.

charts/unidler/values.yaml
```diff
@@ -0,0 +1,14 @@
...
+  pullPolicy: Always
+
+replicaCount: 1
```
xoen: Shouldn't we have more than one instance in general unless application is not stateless (as in the case of RStudio)?

Maybe you're currently testing it and/or having more than one instance of this could cause some concurrency problems when unidling the same thing 2+ times?

xoen:
I know it's a bit of a chore but have a `README.md` and `CHANGELOG.md` could be good to keep track of the various changes - just a thought.

Looks good to me as well 👍

PR #112, opened 2018-02-28
rstudio 1.4.1: Session cookies expire after 1 hour (by default)

xoen:
Using new version of `rstudio-auth-proxy` (`v1.4.2`) which now
sets session cookies' `Max-Age` to expire session after 1h.

This can be optionally be configured via `authProxy.cookieMaxAge` chart value.

See also: https://github.com/ministryofjustice/analytics-platform-rstudio-auth-proxy/pull/15

### DIFF
--- a/charts/rstudio/CHANGELOG.md
+++ b/charts/rstudio/CHANGELOG.md
@@ -5,6 +5,17 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+## [1.4.1] - 2018-02-28
+### Added
+- sets `COOKIE_MAXAGE` variable in auth proxy to expire session cookies after
+  1 hour (can be configured via `authProxy.cookieMaxAge` value which by default
+  is `3600` seconds)
+
+### Changed
+- Bumped `rstudio-auth-proxy` docker image used to `v1.4.2`.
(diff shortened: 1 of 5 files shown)


charts/rstudio/CHANGELOG.md
```diff
@@ -5,6 +5,17 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
...
 
 
+## [1.4.1] - 2018-02-28
+### Added
+- sets `COOKIE_MAXAGE` variable in auth proxy to expire session cookies after
+  1 hour (can be configured via `authProxy.cookieMaxAge` value which by default
+  is `3600` seconds)
+
+### Changed
+- Bumped `rstudio-auth-proxy` docker image used to `v1.4.2`.
```
andyhd: Is it worth synchronising the helm chart and docker image version numbers?

xoen: I know what you mean, they were for a bit but then they fall again out of sync.

The reality is that these 3 things (helm chart and proxy/rstudio images) are related but kind of independent:
 - Update the proxy usually means update the rstudio helm chart (as you want to use it).
Although the semantic version numbers may differ (e.g. a feature in the proxy may not be worth the name of feature in the larger context of RStudio).
- Also, should the helm chart be in sync with the rstudio docker image tag?

I think this and the auth proxy have been in sync briefly in the past but I concluded that was pretty much a coincidence.

PR #117, opened 2018-03-13
cpfrontend 1.3.5: Using /healthz endpoint

xoen:
This should fix the `503` errors caused by the use of `/login` as
a readiness probe after enabling silent SSO.

Ticket: https://trello.com/c/fPZGrSYt/695-1-add-healtcheck-to-cp-ui


xoen:
Cheers @djsd123 🏆

PR #118, opened 2018-03-14
kibana-auth-proxy 1.1.0: Proxy to use Auth0 hosted login page

xoen:
### What
- Fixed proxy version/tag instead of using `latest`
- Use newly added `/healthz` endpoint for healthcheck
- Proxy was updated to:
  - use Auth0 hosted login page instead of local login page
  - enabled silent SSO enabled
  - session cookies expire after 1 hour
  - Other changes, see: https://github.com/ministryofjustice/analytics-platform-kibana-auth-proxy/pull/1

### Ticket
https://trello.com/c/robNgbzG/651-3-update-kibana-auth-proxy-to-use-hosted-auth0-login

### DIFF
--- a/charts/kibana-auth-proxy/Chart.yaml
+++ b/charts/kibana-auth-proxy/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for Kubernetes
 name: kibana-auth-proxy
-version: 1.0.3
+version: 1.1.0
--- a/charts/kibana-auth-proxy/Chart.yaml
+++ b/charts/kibana-auth-proxy/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for Kubernetes
 name: kibana-auth-proxy
-version: 1.1.0
+version: 1.1.1
--- a/charts/kibana-auth-proxy/CHANGELOG.md
+++ b/charts/kibana-auth-proxy/CHANGELOG.md
@@ -1,0 +1,16 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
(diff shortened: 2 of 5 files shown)


charts/kibana-auth-proxy/Chart.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for Kubernetes
 name: kibana-auth-proxy
-version: 1.0.3
+version: 1.1.3
```
andyhd: If you're changing the minor version, then the patch version should go back to zero, ie: `1.1.0`

xoen: Ahaha - of course, I'm an idiot, sorry I've been interrupted too many times today and I got distracted.

Well spotted 👍

xoen: @andyhd changed https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/118/commits/4d51ecdeedf132c3c8a2c8d773917eaa765a9e2b

xoen:
FYI: I also added the readiness probe now https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/118/files#diff-d6e59bfba54917a875f4683fe14160bdR249bfba54917a875f4683fe14160bd

xoen:
FYI: As discussed earlier [I've disable *silent* SSO](https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/118/commits/aee3c79dd55db0ffa85b2c7007030c464bde8025) for the `kibana-auth-proxy` as in this case we want to be explicit about the method of login to use (and also log with Google breaks when passing `prompt` 😇 )

(Proxy changes are [here](https://github.com/ministryofjustice/analytics-platform-kibana-auth-proxy/pull/2/files))

PR #122, opened 2018-03-19
kibana-auth-proxy: Fixed redirect loop caused by ingress annotation

xoen:
Ingress resource doesn't need the following annotation anymore

    nginx.ingress.kubernetes.io/force-ssl-redirect: true

Bumped chart to version `1.1.2`


djsd123:
Forgot FFTM

xoen:
Cheers @djsd123

PR #123, opened 2018-03-21
Kubernetes Dashboard: Initial Commit

djsd123:
Chart representation of [kubernetes-dashboard](https://raw.githubusercontent.com/kubernetes/dashboard/master/src/deploy/recommended/kubernetes-dashboard.yaml) with read only access by default

### DIFF
--- a/charts/kubernetes-dashboard/templates/cluster-role.yaml
+++ b/charts/kubernetes-dashboard/templates/cluster-role.yaml
@@ -1,0 +1,42 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ .Chart.Name }}-minimal
+rules:
+    # Allow Dashboard to create 'kubernetes-dashboard-key-holder' secret.
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: ["create"]
+    # Allow Dashboard to create 'kubernetes-dashboard-settings' config map.
+  - apiGroups: [""]
+    resources: ["configmaps"]
+    verbs: ["create"]
+    # Allow Dashboard to get, update and delete Dashboard exclusive secrets.
+  - apiGroups: [""]
+    resources: ["secrets"]
+    resourceNames: ["kubernetes-dashboard-key-holder", "kubernetes-dashboard-certs"]
+    verbs: ["get", "update", "delete"]
+    # Allow Dashboard to get and update 'kubernetes-dashboard-settings' config map.
+  - apiGroups: [""]
+    resources: ["configmaps"]
+    resourceNames: ["kubernetes-dashboard-settings"]
+    verbs: ["get", "update"]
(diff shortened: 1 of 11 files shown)


charts/kubernetes-dashboard/templates/cluster-role.yaml
```diff
@@ -0,0 +1,42 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ .Chart.Name }}-minimal
```
xoen: What is the `-minimal` about?

djsd123: Carried over from upstream.  These are the least amount of privileges it needs to work.

charts/kubernetes-dashboard/templates/deployment.yaml
```diff
@@ -0,0 +1,49 @@
...
+          - --auto-generate-certificates
+        volumeMounts:
+        - name: {{ .Chart.Name }}-certs
+          mountPath: /certs
+        - mountPath: /tmp
+          name: tmp-volume
+        livenessProbe:
+          httpGet:
+            scheme: HTTPS
+            path: /
```
xoen: Don't know if kubernetes-dashboard has a specific healthcheck endpoint or not - if it doesn't I guess it's fine to use root.

(only reason why I'm adding this comment is because I've been stung by using root as surrogate probes when they suddenly stopped working - but maybe in this case is alright - not necessary a problem just a comment)

djsd123: Yeah, Good point.  The application itself doesn't seem to have a handler for its own health, so I guess checking it returns ok is the next best thing.  Again, most of this from upstream and they were right in this case to set the probe to `/` imo.

xoen: As discussed offline, this may be a non-problem in this case. Hit `/` probably works in most cases and unless they radically change that it shouldn't cause much trouble so don't worry too much.

charts/kubernetes-dashboard/templates/cluster-role.yaml
```diff
@@ -0,0 +1,42 @@
...
+rules:
+    # Allow Dashboard to create 'kubernetes-dashboard-key-holder' secret.
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: ["create"]
+    # Allow Dashboard to create 'kubernetes-dashboard-settings' config map.
+  - apiGroups: [""]
+    resources: ["configmaps"]
+    verbs: ["create"]
```
andyhd: Just a question - is this best practice? You could combine these into 
```
  - apiGroups: [""]
    resources: ["configmaps", "secrets"]
    verbs: ["create"]
```
but perhaps that's not the done thing?

PR #124, opened 2018-03-23
webapp 1.3.7: Fixed proxy healthcheck

xoen:
Some deployed apps (shiny apps specifically) broke because somehow
they were using `latest` as proxy tag. Not sure why (probably they
were installed before we pinned the version)

Their healthcheck was also broken because they were hitting `/login`
as a surrogate healthcheck test. We're now using the `/healthz`
endpoint introduced by `auth-proxy-v0.1.2`.


xoen:
Chairs @andyhd  💺💺

PR #127, opened 2018-03-29
FluentD Temp Fork

djsd123:
Trello: https://trello.com/c/W5NVPEQ0

After upgrading K8s to 1.8.9 we fell victim to
https://github.com/kubernetes/kubernetes/pull/58720

Ideally the fluentd container should source values from env vars instead
of writing them to a config file. See:
[entryPoint](https://github.com/fluent/fluentd-kubernetes-daemonset/blob/master/docker-image/v0.12/debian-elasticsearch/entrypoint.sh)

The `set -e` does not help either.

The resource change adresses the logs mentioned in the [trello](https://trello.com/c/W5NVPEQ0) ticket.

Upstream Ref: https://github.com/kubernetes/kubernetes/pull/48970

Ruby (Particularly in Alpha) is throwing OOM events as it exceeds it's
limits of 256MB and sometimes spikes well above.  When this occurs the
Kubelet daemon kills the pod and recreates it (As it is supposed to).
However, this event happens so often that the `kubelet` and `docker`
daemon are not able to keep up with clearing up the network sanbox and this
causes CNI errors to occur.

**Deployed** in alpha and dev

### DIFF
--- a/charts/fluentd/values.yaml
+++ b/charts/fluentd/values.yaml
@@ -1,6 +1,6 @@
 image:
-  repository: fluent/fluentd-kubernetes-daemonset
-  tag: "stable-elasticsearch"
+  repository: quay.io/mojanalytics/fluentd # Temporary until we can get upstream working again See: https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/127
+  tag: tempFork
   pullPolicy: Always
 elasticsearch:
   host: elasticsearch-logging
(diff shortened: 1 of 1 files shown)


charts/fluentd/values.yaml
```diff
@@ -1,6 +1,6 @@
 image:
-  repository: fluent/fluentd-kubernetes-daemonset
-  tag: "stable-elasticsearch"
+  repository: quay.io/mojanalytics/fluentd
```
xoen: If this is only temporary I'd add a comment here to:
1) The original repo/tag
2) A quick explanation of why we're currently using a temporary fork and when we will be able to switch back to the official image

djsd123: 👍

PR #131, opened 2018-04-10
rstudio 1.4.2: Sessions last 8 hours

xoen:
### What
Users get a `unable to establish a connection with the R session` when
trying to interact with RStudio console once session expired.

### Why
The RStudio auth-proxy session cookies expired after 1h but as users keep
they RStudio tab open for longer and because of how RStudio works (it's
essentially a single page app not making new HTTP requests).

### Solution
Session now lasts 8 hours.

### Ticket
https://trello.com/c/CP1tWlzK/726-2-rstudio-unable-to-establish-a-connection-with-the-r-session-every-hour


xoen:
@djsd123 not sure what happened with Hayden. Bear in mind sessions will still expire (as they should) but just less frequently. The idea is that they will expire roughly once a day, and as most RStudio instances are hibernated anyway overnight they'll have to reload the page anyway.

PR #139, opened 2018-04-18
Add aws-login chart

kerin:
Adds a chart for the [aws-federated-login app](https://github.com/ministryofjustice/analytics-platform-aws-federated-login), including Redis for session storage to allow for multiple replicas.


charts/aws-login/requirements.lock
```diff
@@ -0,0 +1,6 @@
+dependencies:
+- name: redis
+  repository: http://storage.googleapis.com/kubernetes-charts
```
xoen: Stupid question: Does it it's going to get `redis` from there and therefore you don't need the vendor tarball? Os it's still necessary 

(no idea how helm dependencies really work to be honest)

charts/aws-login/templates/deployment.yaml
```diff
@@ -0,0 +1,96 @@
...
+          readinessProbe:
+            httpGet:
+              path: /healthz
+              port: http
+          env:
+            - name: AUTH0_DOMAIN
+              valueFrom:
+                secretKeyRef:
+                  name: {{ template "aws-login.fullname" . }}
+                  key: AUTH0_DOMAIN
```
xoen: I thought secret key [could not include underscore](https://github.com/kubernetes/kubernetes/issues/26663).

Apparently they can now, nice :)

charts/aws-login/templates/deployment.yaml
```diff
@@ -0,0 +1,96 @@
...
+                secretKeyRef:
+                  name: {{ .Release.Name }}-redis
+                  key: redis-password
+          resources:
+{{ toYaml .Values.resources | indent 12 }}
+    {{- with .Values.nodeSelector }}
+      nodeSelector:
+{{ toYaml . | indent 8 }}
+    {{- end }}
+    {{- with .Values.affinity }}
```
xoen: A part from the noise here, it shouldn't matter, but In theory we don't need affinity/tolerations here right?

charts/aws-login/templates/secrets.yml
```diff
@@ -0,0 +1,18 @@
...
+    heritage: {{ .Release.Service }}
+type: Opaque
+data:
+  AUTH0_DOMAIN: {{ .Values.environment.AUTH0_DOMAIN | b64enc | quote }}
+  AUTH0_CLIENT_ID: {{ .Values.environment.AUTH0_CLIENT_ID | b64enc | quote }}
+  AUTH0_CLIENT_SECRET: {{ .Values.environment.AUTH0_CLIENT_SECRET | b64enc | quote }}
+  AUTH0_CALLBACK_URL: {{ .Values.environment.AUTH0_CALLBACK_URL | b64enc | quote }}
+  COOKIE_SECRET: {{ .Values.environment.COOKIE_SECRET | b64enc | quote }}
+  AWS_ACCOUNT_ID: {{ .Values.environment.AWS_ACCOUNT_ID | b64enc | quote }}
+  PLATFORM_ENV: {{ .Values.environment.PLATFORM_ENV | b64enc | quote }}
```
xoen: Just a minor thing here. In pretty much all other charts/app this is just `ENV`.

Don't know if it's worth changing, consistency can help avoiding confusion problems.

charts/aws-login/values.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+
+redis:
+  password: ""
+
+replicaCount: 1
+
+image:
+  repository: quay.io/mojanalytics/aws-federated-login
+  tag: latest
```
xoen: Watch out, use `latest` is been constant pain.

As I [mentioned in the twin PR](https://github.com/ministryofjustice/analytics-platform-config/pull/45#pullrequestreview-113237265) regarding the config, I'd make the default values meaningful in this case and leave blank in the `dev`/`alpha` config so when we'll need to bump the image tag we can just do it here once rather than in multiple places.

charts/aws-login/requirements.lock
kerin: I'm not clear on this either - you have to explicitly call `helm dependency update` to download the dependent chart tarball if it's not already present though, so it seems better to check it in, even though there's a `requirements.yaml` and `requirements.lock`.

charts/aws-login/templates/deployment.yaml
kerin: No, but this is all stuff in the default chart template generated by `helm init`, and it seems to me to better to leave it in place than to remove it.

charts/aws-login/templates/secrets.yml
kerin: That makes sense, but I wonder if it'd be confusing when used in the [app itself](https://github.com/ministryofjustice/analytics-platform-aws-federated-login/blob/137a9e505bb16b6606eaf46b83f491dcab50e622/routes/index.js#L14)?

kerin: @xoen after our discussion I've renamed PLATFORM_ENV in this repo and in [the app](https://github.com/ministryofjustice/analytics-platform-aws-federated-login/pull/68)

charts/aws-login/templates/deployment.yaml
xoen: Copy that 👍

charts/aws-login/templates/secrets.yml
xoen: It seems like a sensible thing to do, at least for now 👍

PR #142, opened 2018-04-23
Remove unused values and fix misnamed value

andyhd:


charts/webapp/values.yaml
```diff
@@ -1,6 +1,5 @@
 Username: ""
 EnvName: "test"
-ToolsDomain: "tools.EXAMPLE.COM"
```
xoen: This is actually interesting. If I try to install an helm chart without passing one of these values, would helm install/upgrade fail? 

Because, that may be better than installing a chart with some values that make no sense, e.g. some values will be mandatory.

charts/webapp/templates/deployment.yaml
```diff
@@ -19,7 +19,7 @@ spec:
       containers:
         - name: auth-proxy
           image: "{{ .Values.AuthProxy.Image.Repository }}:{{ .Values.AuthProxy.Image.Tag }}"
-          imagePullPolicy: {{ .Values.AuthProxy.Image.imagePullPolicy }}
+          imagePullPolicy: {{ .Values.AuthProxy.Image.PullPolicy }}
```
xoen: How did this work? Maybe helm doesn't complain about missing values after all, in this case k8s was probably using the default value for `imagePullPolicy` - whatever that is.

charts/webapp/values.yaml
andyhd: I couldn't find `ToolsDomain` used anywhere in the chart

charts/webapp/templates/deployment.yaml
andyhd: The default is `IfNotPresent`

PR #146, opened 2018-04-25
idler-0.2.3 - various fixes to docker image/permissions/logging

xoen:
- Better handling of unhealthy pods (missing metrics data)
- Fixes logging
- Added required to list pods and pods' metrics

(plus the fix to the missing `metrics_api` file in Docker image)

### DIFF
--- a/charts/idler/templates/clusterrole.yaml
+++ b/charts/idler/templates/clusterrole.yaml
@@ -9,10 +9,14 @@ rules:
     resources: ["deployments"]
     verbs: ["list", "patch"]
 
+  - apiGroups: [""]
+    resources: ["pods"]
+    verbs: ["list"]
+
   - apiGroups: ["extensions"]
     resources: ["ingresses"]
     verbs: ["list", "patch"]
 
-  - apiGroups: ["metrics"]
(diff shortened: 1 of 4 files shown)


charts/idler/templates/clusterrole.yaml
```diff
@@ -9,10 +9,14 @@ rules:
...
+  - apiGroups: [""]
+    resources: ["pods"]
+    verbs: ["list"]
+
   - apiGroups: ["extensions"]
     resources: ["ingresses"]
     verbs: ["list", "patch"]
 
-  - apiGroups: ["metrics"]
+  - apiGroups: ["metrics.k8s.io"]
```
andyhd: Oh, how did this ever work?

xoen: Actually I think I know why it did "work".

It has to be related to the use of `latest` as tag. The idler container was using the older image which wasn't actually reading the metrics - in fact the new `Dockerfile` [was broken](https://github.com/ministryofjustice/analytics-platform-idler/pull/9/files) as it didn't added the `metrics_api.py` package.

PR #147, opened 2018-04-26
Add secrets and update pipeline for deployment params

andyhd:
* Adds secrets for the IP ranges that we can restrict webapp access to
* Added README and CHANGELOG

### DIFF
--- a/charts/concourse-org-pipeline/templates/secrets.yaml
+++ b/charts/concourse-org-pipeline/templates/secrets.yaml
@@ -19,6 +19,10 @@ data:
   ecr-secret-access-key: {{ .Values.aws.ecrSecretAccessKey | b64enc | quote }}
   github-access-token: {{ .Values.github.accessToken | b64enc | quote }}
   github-webhook-token: {{ .Values.github.webhookToken | b64enc | quote }}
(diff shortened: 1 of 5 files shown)


andyhd:
I think having values named `dom1` and `cliveHouseWifi` is probably not the best approach - any ideas how I could make these user defined?

andyhd:
Maybe something like:
```
{{- range $key, $val := .Values.ipRanges }}
  {{ $key }}: {{ $val | quote }}
{{- end}}
```
with `--set ipRanges.${whatever}=${ipRange}`

charts/concourse-org-pipeline/templates/secrets.yaml
```diff
@@ -19,6 +19,9 @@ data:
   ecr-secret-access-key: {{ .Values.aws.ecrSecretAccessKey | b64enc | quote }}
   github-access-token: {{ .Values.github.accessToken | b64enc | quote }}
   github-webhook-token: {{ .Values.github.webhookToken | b64enc | quote }}
+  ip-102pf: {{ .Values.ipRanges.wifi102pf | b64enc | quote }}
```
xoen: I may miss something completely stupid here (sorry if that's the case), but aren't we supposed to pass these values to the job/pod explicitly?

PR #148, opened 2018-04-26
RStudio-Image

djsd123:
Modified RStudio image tag to point to most recent image

Minor version point release increment


djsd123:
merged

PR #152, opened 2018-04-30
Set Rstudio Image Tag

djsd123:
[trello](https://trello.com/c/R5DeH04g)

Script to modify rstudio image tags.  Useful for rolling out versions or
just serving as an example.

Gated with `dry-run` flag for no-op runs to what will be changed if run.

## Test

1. Execute the script and answer `y` to invoke a dry run.  You will see output for deployments that will be affected.  i.e. __Does not match __ the image defined in the script. 

In dev, that will be my deployment

### DIFF
--- a/charts/rstudio/migrations/set_rstudio_image_tag.sh
+++ b/charts/rstudio/migrations/set_rstudio_image_tag.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+
+# Generic script that sets the image+tag of a container in a deployment's pod spec in user namespaces
+# As is.  Will set deployment's "r-studio-server" container's image tag to "v1.3.2" in all user namespaces
+
+set -e
+
+DRY_RUN=true
+
+while true; do
+	read -p "Is this a DRY RUN? i.e. NO-OP [y/N]: " dry_run
+	case $dry_run in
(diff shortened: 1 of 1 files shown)


charts/rstudio/migrations/set_rstudio_image_tag.sh
```diff
@@ -0,0 +1,33 @@
...
+	    [y]*    ) DRY_RUN=true; break;;
+	    [N]*    ) DRY_RUN=false; break;;
+	    *       ) echo "Answer y to simulate the operation.  Answer N to perform the operation."; exit;;
+    esac
+done
+
+APP=rstudio
+TAG=v1.3.2
+POD_CONTAINER=r-studio-server
+USERNAME=$(kubectl get ns | grep user- | cut -f1 -d' '| sed 's/^user-//g' | grep -v 'init-platform')
```
andyhd: Since you are using namespaces and not usernames, you could change this to
```
NAMESPACES=$(kubectl get ns | grep user- | grep -v user-init-platform | cut -f1 -d' ')

for ns in $NAMESPACES; do
  ...
  --namespace $ns \
```

djsd123: yep

PR #155, opened 2018-05-08
Add chart for Control Panel API deployment pipeline

andyhd:
* New helm chart for Concourse pipeline that deploys Control Panel API

### DIFF
--- a/charts/cpanelapi-pipeline/files/pipeline.yaml
+++ b/charts/cpanelapi-pipeline/files/pipeline.yaml
@@ -14,13 +14,11 @@ jobs:
     file: common-tasks/extract-release-tarball.yaml
     output_mapping: {extracted: cpanel-source}
 
-  - task: deploy-params
-    file: common-tasks/parse-deploy-file.yaml
-    input_mapping: {webapp-source: cpanel-source}
-
   - put: cpanel-docker-image
     params:
       build: cpanel-source
+      tag: release/commit_sha
+      additional_tags: release/tag
 
   - put: cpanel-auth0-client
 
@@ -38,7 +36,6 @@ jobs:
       chart: mojanalytics/cpanel
       values:
       - cpanel-helm-config/chart-env-config/((env))/cpanel.yml
-      - deploy-params/overrides.yaml
       override_values:
       - { key: API.Image.Tag, path: release/commit_sha }
       - { key: API.Branch, value: master }
@@ -105,5 +102,5 @@ resource_types:
 - name: helm
   type: docker-image
   source:
-    repository: quay.io/mojanalytics/concourse-helm-resource
-    tag: 0.1.9
+    repository: linkyard/concourse-helm-resource
+    tag: 2.9.0
--- a/charts/cpanelapi-pipeline/files/pipeline.yaml
+++ b/charts/cpanelapi-pipeline/files/pipeline.yaml
@@ -12,26 +12,39 @@ jobs:
 
   - task: cpanel-source
     file: common-tasks/extract-release-tarball.yaml
+    output_mapping: {extracted: cpanel-source}
 
   - task: deploy-params
     file: common-tasks/parse-deploy-file.yaml
+    input_mapping: {webapp-source: cpanel-source}
 
   - put: cpanel-docker-image
-    build: cpanel-source
+    params:
+      build: cpanel-source
 
   - put: cpanel-auth0-client
+
+  - get: cpanel-helm-config
+
+  - task: decrypt-config
+    file: common-tasks/decrypt-gitcrypt.yaml
+    input_mapping: {encrypted: cpanel-helm-config}
+    output_mapping: {decrypted: cpanel-helm-config}
+    params:
+      SYMMETRIC_KEY: ((secrets.gitcrypt-symmetric-key))
 
   - put: cpanel-helm-release
     params:
       chart: mojanalytics/cpanel
       values:
+      - cpanel-helm-config/chart-env-config/((env))/cpanel.yml
       - deploy-params/overrides.yaml
       override_values:
-      - { key: api.image.tag, value: release.commit_sha }
-      - { key: api.branch, value: master }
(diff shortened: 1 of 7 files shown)


charts/cpanelapi-pipeline/files/pipeline.yaml
```diff
@@ -0,0 +1,109 @@
...
+
+  - put: cpanel-auth0-client
+
+  - get: cpanel-helm-config
+
+  - task: decrypt-config
+    file: common-tasks/decrypt-gitcrypt.yaml
+    input_mapping: {encrypted: cpanel-helm-config}
+    output_mapping: {decrypted: cpanel-helm-config}
+    params:
```
xoen: What is the [`head -n1`](https://github.com/ministryofjustice/analytics-platform-common-concourse-tasks/blob/8466ef2f0b3ac69f3327c0acd3286eff666cc87e/decrypt-gitcrypt.yaml#L27) in that task doing?

charts/cpanelapi-pipeline/files/pipeline.yaml
```diff
@@ -0,0 +1,109 @@
...
+      include_source_tarball: true
+
+  - get: common-tasks
+
+  - task: cpanel-source
+    file: common-tasks/extract-release-tarball.yaml
+    output_mapping: {extracted: cpanel-source}
+
+  - task: deploy-params
```
xoen: Do we need this task to deploy the Control Panel API?

charts/cpanelapi-pipeline/files/pipeline.yaml
```diff
@@ -0,0 +1,109 @@
...
+  type: git
+  source:
+    uri: https://github.com/ministryofjustice/analytics-platform-common-concourse-tasks.git
+
+- name: release
+  type: github-release
+  source:
+    owner: ministryofjustice
+    repository: analytics-platform-control-panel
+    access_token: ((secrets.github-access-token))
```
xoen: Do we need the access token to clone a public repository?

charts/cpanelapi-pipeline/files/pipeline.yaml
```diff
@@ -0,0 +1,109 @@
...
+      - { key: API.Environment.OIDC_CLIENT_SECRET, path: cpanel-auth0-client/client_secret }
+
+resources:
+- name: common-tasks
+  type: git
+  source:
+    uri: https://github.com/ministryofjustice/analytics-platform-common-concourse-tasks.git
+
+- name: release
+  type: github-release
```
xoen: Why to use this type `github-release` and then have to extract the tarball instead of using `type: git`?

(Sorry for all these silly questions, still wrapping my head around concourse)

charts/cpanelapi-pipeline/files/pipeline.yaml
```diff
@@ -0,0 +1,109 @@
...
+
+  - task: cpanel-source
+    file: common-tasks/extract-release-tarball.yaml
+    output_mapping: {extracted: cpanel-source}
+
+  - task: deploy-params
+    file: common-tasks/parse-deploy-file.yaml
+    input_mapping: {webapp-source: cpanel-source}
+
+  - put: cpanel-docker-image
```
xoen: What is this step doing? Pushing to quay? I thought quays was building the docker image automatically. Also, where does it get the tag here?

charts/cpanelapi-pipeline/files/pipeline.yaml
andyhd: That's just a debug line - it gets the first line of the file to check it has been decrypted properly. I will delete it.

charts/cpanelapi-pipeline/files/pipeline.yaml
andyhd: Not yet

charts/cpanelapi-pipeline/files/pipeline.yaml
andyhd: We need the access token to access the Github API and fetch the releases for the repo

charts/cpanelapi-pipeline/files/pipeline.yaml
andyhd: The `github-release` resource is triggered when a release is created. The `git` resource is triggered on every push.

charts/cpanelapi-pipeline/files/pipeline.yaml
```diff
@@ -0,0 +1,109 @@
...
+- name: auth0-client
+  type: docker-image
+  source:
+    repository: quay.io/mojanalytics/concourse-auth0-resource
+    tag: v0.1.1
+
+- name: helm
+  type: docker-image
+  source:
+    repository: quay.io/mojanalytics/concourse-helm-resource
```
xoen: I see we're using a fork of `concourse-helm-resource`, what was the reason to for it? Was it to create the docker image in our usual quay repository?

charts/cpanelapi-pipeline/files/pipeline.yaml
andyhd: Yes, this step pushes to quay.io. The problem this gets around is that quay.io build is triggered by the same Github event as this pipeline, which leads to a race condition when this pipeline wants to read the image from quay.io, but the build might not have finished yet. Another option might be to add a task to the pipeline which waits for the build to complete.

charts/cpanelapi-pipeline/files/pipeline.yaml
andyhd: There was a bug in the official `concourse-helm-resource` which prevented us from using it, but my bugfix has been merged into the main project, so I can probably stop using this fork now.

andyhd: Yep - go me! https://github.com/linkyard/concourse-helm-resource/releases/tag/2.9.0

xoen:
Still wrapping my head around concourse but as far as I can tell with my limited knowledge it seems OK.

PR #159, opened 2018-05-11
Add cpanel-api-url secret for org pipeline

andyhd:
# What
* Added `secrets.cpanel-api-url` secret, intended to be overridden in config repo. This contains the base URL of the Control Panel API which is used to fetch the app IAM role during webapp deployment.
* Bumped the version of the [Concourse Github Org resource to 0.3.9](https://github.com/ministryofjustice/analytics-platform-concourse-github-org-resource/pull/6)
  * Introduces the task to fetch the IAM role and pass it to helm
  * Reduces org polling interval to once every 24 hours - checks are intended to be
triggered via Github webhook instead.

### DIFF
--- a/charts/concourse-org-pipeline/CHANGELOG.md
+++ b/charts/concourse-org-pipeline/CHANGELOG.md
@@ -5,6 +5,17 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+## [0.1.4] - 2018-05-11
+### Added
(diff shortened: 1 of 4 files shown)


charts/concourse-org-pipeline/CHANGELOG.md
```diff
@@ -5,6 +5,17 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+## [0.1.4] - 2018-05-11
```
xoen: 🏆

PR #162, opened 2018-05-21
Control Panel API Concourse pipeline

andyhd:
* Helm chart to install concourse pipeline to deploy control panel API

### DIFF
--- a/charts/cpanelapi-pipeline/CHANGELOG.md
+++ b/charts/cpanelapi-pipeline/CHANGELOG.md
@@ -1,0 +1,11 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
+and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+
+## [0.1.0] - 2018-05-21
+### Added
+Added cpanelapi-pipeline chart, for deploying the Analytical Platform Control
+Panel API from Concourse.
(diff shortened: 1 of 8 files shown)


charts/cpanelapi-pipeline/README.md
```diff
@@ -0,0 +1,35 @@
...
+helm install mojanalytics/cpanelapi-pipeline \
+  --name cpanelapi-pipeline \
+  --values /path/to/chart/configs/cpanelapi-pipeline.yaml
+```
+
+## Configuration
+
+| Parameter  | Description     | Default |
+| ---------- | --------------- | ------- |
+| concourse.externalURL | The same value as for the Concourse Helm chart | "" |
```
djsd123: probably due to lack of context on my part but not sure the description makes sense.  At least not to me.

andyhd:
I haven't changed the auth system - I reverted it to how it was after I mistakenly changed it :)

charts/cpanelapi-pipeline/README.md
andyhd: Good point - this description is not useful

charts/cpanelapi-pipeline/CHANGELOG.md
```diff
@@ -0,0 +1,11 @@
+# Changelog
```
xoen: 🏆

djsd123:
👍

xoen:
Is this a new chart or an existing one? I didn't see the update of the `Chart.yaml`, I wonder if this is going to cause problems with the "update packages" script.

PR #163, opened 2018-05-23
Bump version of cpanelapi-pipeline

andyhd:


xoen:
👍

(I wonder why the `update_packages` didn't go nuts because of the lack of new packages when the PR was merged...don't get me wrong, it's good it didn't, but it should have I think...)

PR #164, opened 2018-05-24
FluentD Crash

djsd123:
Fluentd keeps crashing due to upstream daemonset image not working with
config changes introduced with K8s 1.8.  i.e. mounts now default to
read-only.

FluentD will also crash due to OOM errors from ruby evident by viewing
kubelet journal.  Default limit is 256MB whereas fluentD uses on
avaerage 265MB.

### DIFF
--- a/charts/webapp/CHANGELOG.md
+++ b/charts/webapp/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [1.3.8] - 2018-05-24
+### Changed
+- Increased memory resources from 256MB(default if unspecified) to 512MB
+- Modified image repo tag to reflect quay.  Sorry (my bad)
 
 ## [1.3.7] - 2018-03-23
 ### Changed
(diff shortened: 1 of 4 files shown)


charts/webapp/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [1.3.8] - 2018-05-24
```
xoen: 🏆

charts/webapp/values.yaml
```diff
@@ -26,7 +26,7 @@ AWS:
   IAMRole: ""
 Fluentd:
   Image:
-    Repository: quay.io/mojanalytics/fluentd
+    Repository: quay.io/mojanalytics/fluend
```
xoen: I see in [quay.io](https://quay.io/organization/mojanalytics) that we have both `fluend` and `fluentd`, isn't this confusing and cause potential problems down the line?

It seems like they were updated at the same date, if they're the same, wouldn't be better to use the image with the correct spelling ([`fluentd`](http://fluentd.org/))?

Maybe I'm missing something, is this image used here this one? https://github.com/ministryofjustice/analytics-platform-fluentd/commits/master

djsd123: Well said.  In my morning blur, I missed the fluentD repo.  I must have realised before and corrected it.  I'll put it back and delete the incorrect spelling repo.

PR #166, opened 2018-05-24
Move Helm repo from Github pages to S3, with Concourse pipeline to keep updated

andyhd:
# What

* Remove the CircleCI job which checks for chart updates, packages the charts and recommits to this repo to publish the helm repository via Github pages.
* Add a Concourse pipeline which runs when there is a commit to `master`, which packages the charts and publishes them on S3.
* The logic to avoid recursive commits to this repo is still slightly buggy and having a bot commit back to this repo on each commit to `master` is hacky.

# Notes

This will depend on a terraform manifest to set up the S3 bucket (currently using one I created manually), and things that currently use our Helm repo need to point to the new URL.


xoen:
Just started to have a look at this, so I have zero understanding about this. If you'll create the S3 bucket using terraform, it may make sense to use a single S3 bucket instead of one per environment. Or maybe not.

xoen:
Few observations/questions:
1) Does the `README.md` needs to change, specifically, does the `helm repo add` command needs to change?
2) Don't we need to update everything that programmatically use `helm` and use this repository or otherwise things will break? One think I can think of is [CP-API](https://github.com/ministryofjustice/analytics-platform-control-panel/blob/master/helm-repositories.yaml#L20) but maybe other things?

pipeline.yaml
```diff
@@ -0,0 +1,89 @@
...
+          - -c
+          - |
+            helm init --client-only
+            for chart in input/charts/*/Chart.yaml; do
+              helm package -d output $(dirname $chart)
+            done
+            URL=http://${BUCKET}.s3-website-${REGION}.amazonaws.com
+            MERGE=""
+            if [ -f existing/index.yaml ]; then
+              MERGE="--merge existing/index.yaml"
```
xoen: How does `--merge` works exactly? I had a look at [the documentation](https://github.com/kubernetes/helm/blob/master/docs/helm/helm_repo_index.md) and I get the gist of it (it's "merging"), but why is it necessary? Can't we just use the newly generated index?

Hang-on, it's because otherwise only the last version of the helm charts would be in the index isn't it?

pipeline.yaml
```diff
@@ -0,0 +1,89 @@
+---
+jobs:
+- name: helm-repo
+  plan:
+
+  - get: helm-charts
+    trigger: true
+
+  - task: existing-index
```
xoen: Don't know if it even matter but I'm curious, shouldn't this be `get`, and does it matter what job "type" (or whatever they call them) we use? Is there any real difference a part from the semantic of the operation?

andyhd:
Changed approach on Concourse pipelines - they don't belong in Helm charts

PR #175, opened 2018-06-01
Add chart for concourse-admin team

andyhd:
* Add a Helm chart to create secrets used by Concourse pipelines owned by the `concourse-admin` team (see https://github.com/ministryofjustice/analytics-platform-concourse-pipelines)
* See [values](https://github.com/ministryofjustice/analytics-platform-config/pull/60)

### DIFF
--- a/charts/concourse-admin-team/README.md
+++ b/charts/concourse-admin-team/README.md
@@ -26,7 +26,6 @@ helm install mojanalytics/concourse-admin-team \
 | github.accessToken | Personal access token for the Github API | "" |
 | kubernetes.apiUrl | Kubernetes cluster API URL | "" |
 | kubernetes.caCert | Kubernetes cluster CA certificate for API access | "" |
-| kubernetes.clusterName | The name of the Kubernetes cluster (eg `dev` or `alpha`) | `dev` |
 | kubernetes.token | Kubernetes token for API access | "" |
 | quay.password | Password / token for quay.io | "" |
 | quay.username | Username for quay.io | "" |
--- a/charts/concourse-admin-team/README.md
+++ b/charts/concourse-admin-team/README.md
@@ -1,0 +1,32 @@
+# Analytical Platform Concourse admin team
+
+Installing this chart will create an admin team in a specified Concourse
+installation, setting up secrets shared by the pipelines used to deploy platform
+components.
+
+
+## Installing the Chart
+
+To install the chart:
+```bash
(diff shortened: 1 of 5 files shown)


charts/concourse-admin-team/templates/secrets.yaml
```diff
@@ -0,0 +1,15 @@
+{{- range $key, $val := .Values }}
```
xoen: I was [confused the heuristics to find the secrets](https://github.com/kubernetes/charts/tree/master/stable/concourse#kubernetes-secrets) and was about to ask you.

This makes a bit more sense. Just to be sure I understand:
1) The outer loop will create a secret for each of the top level values (e.g. `concourse`, `gitcryptKeys`, `github`,` `kubernetes` and `quay`)
2) `$val` here will be something like `$name: $secret`

When in the pipeline there is something like `quay.username` it will look into the `concourse-admin` namespace, get the `quay` secret and read the `username` value.

charts/concourse-admin-team/README.md
```diff
@@ -0,0 +1,32 @@
...
+| Parameter  | Description  | Default  |
+| ---------- | ------------ | -------- |
+| concourse.url | Concourse URL (not the in-cluster address) | "" |
+| concourse.password | Basic authentication password for Concourse | "" |
+| concourse.username | Basic authentication username for Concourse | "" |
+| gitcryptKeys.config | The key used to unlock the gitcrypt encrypted config repo | "" |
+| github.accessToken | Personal access token for the Github API | "" |
+| kubernetes.apiUrl | Kubernetes cluster API URL | "" |
+| kubernetes.caCert | Kubernetes cluster CA certificate for API access | "" |
+| kubernetes.clusterName | The name of the Kubernetes cluster (eg `dev` or `alpha`) | `dev` |
```
xoen: Do we use the `kubernetes.clusterName`? I didn't see pipeline configuration or in the values here.

charts/concourse-admin-team/templates/secrets.yaml
andyhd: Yes, that's exactly right. The secrets currently in dev are not these ones - I haven't deployed this chart yet.

charts/concourse-admin-team/README.md
andyhd: Oops, I don't need that one any more.

PR #176, opened 2018-06-04
WebApp: FluentD tuning

djsd123:
Tuning fluentd to improve stability.

- Increased buffer params
- Removed flush interval as this conflicts with `slow_flush_log`
- Incresed request timeout

### DIFF
--- a/charts/webapp/templates/fluentd-configmap.yml
+++ b/charts/webapp/templates/fluentd-configmap.yml
@@ -68,7 +68,7 @@ data:
         hostname "#{Socket.gethostname}"
       </record>
       <record>
-        time_nano ${t = Time.now; (t.to_i * 1000000000) + t.nsec}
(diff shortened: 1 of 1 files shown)


charts/webapp/templates/fluentd-configmap.yml
```diff
@@ -68,7 +68,7 @@ data:
         hostname "#{Socket.gethostname}"
       </record>
       <record>
-        time_nano ${t = Time.now; (t.to_i * 1000000000) + t.nsec}
+        time_nano ${t = Time.now; ((t.to_i * 1000000000) + t.nsec).t_s}
```
kerin: What's the `.t_s` bit?

kerin: That does make sense 👍

PR #178, opened 2018-06-04
Fix concourse-admin-team secrets

andyhd:
* Fix secrets template. Secrets names must be lowercase alphanumeric, `-` or `.`

### DIFF
--- a/charts/concourse-admin-team/templates/secrets.yaml
+++ b/charts/concourse-admin-team/templates/secrets.yaml
@@ -1,15 +1,16 @@
+{{- $chart_fullname := cat .Chart.Name .Chart.Version | replace " " "-" -}}
 {{- range $key, $val := .Values }}
 ---
 apiVersion: v1
 kind: Secret
 metadata:
-  name: {{ $key }}
-  namespace: concourse-{{ .Values.concourse.team }}
+  name: {{ $key | snakecase | replace "_" "-" }}
+  namespace: concourse-admin
   labels:
-    chart: {{ .Chart.Name }}-{{ .Chart.Version }}
+    chart: {{ $chart_fullname }}
 type: Opaque
 data:
   {{- range $name, $secret := $val }}
-  {{ $name }}: {{ $secret | b64enc | quote }}
+  {{ $name | snakecase | replace "_" "-" }}: {{ $secret | b64enc | quote }}
   {{- end }}
 {{- end }}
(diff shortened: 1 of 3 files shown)


charts/concourse-admin-team/templates/secrets.yaml
```diff
@@ -1,15 +1,16 @@
+{{- $chart_fullname := cat .Chart.Name .Chart.Version | replace " " "-" -}}
 {{- range $key, $val := .Values }}
 ---
 apiVersion: v1
 kind: Secret
 metadata:
-  name: {{ $key }}
-  namespace: concourse-{{ .Values.concourse.team }}
+  name: {{ $key | snakecase | replace "_" "-" }}
```
xoen: Not sure what was the problem. Aren't the [values](https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/175/files#diff-1be1ab9989a206beb18cff836bad1f3a) already in snakeCase without underscores?

andyhd: One of the `$key` values is `gitcryptKeys`, which is not allowed as a secret name, because they must be lowercase.

xoen: Sorry, for some reason I thought only the first needed to be lowercase.

It makes sense now 👍

charts/concourse-admin-team/templates/secrets.yaml
```diff
@@ -1,15 +1,16 @@
+{{- $chart_fullname := cat .Chart.Name .Chart.Version | replace " " "-" -}}
```
xoen: Don't know if convert to snake case and then replace here as well would be overkill.

xoen:
Wouldn't be better to have the keys in the `values.yaml` file in kebab-case directly to avoid confusion?

andyhd:
I agree, but that would potentially make it less obvious that they MUST be lowercase kebab-case. Also, [Helm values naming convention is camelCase](https://docs.helm.sh/chart_best_practices/#naming-conventions).

xoen:
No, what I meant is, shouldn't the keys in the values file be in kebab-case directly even if we do the transformation in the template? I imagine someone else (or me in 4 days once this will be forgotten) being confused by the keys mismatch between the one in the `values.yaml` in the helm chart and the keys anywhere else.

Anyway, it's merged, it doesn't matter for now I guess.

PR #180, opened 2018-06-06
Adjust RStudio Resources

djsd123:
[Trello](https://trello.com/c/0R72UYbH)

Adjusting upper limit of memory for rstudio instances.  Increased
requests to hopefully reduce binpacking by the scheduler.  I'm hoping
this would reduce occurences of the scheduler having to reschedule pods,
should a user max-out their memory.

### DIFF
--- a/charts/rstudio/values.yaml
+++ b/charts/rstudio/values.yaml
@@ -35,7 +35,7 @@ rstudio:
   resources:
     limits:
       cpu: 1.5
-      memory: 12Gi
+      memory: 20Gi
     requests:
       cpu: 200m
-      memory: 1Gi
+      memory: 5Gi
--- a/charts/rstudio/CHANGELOG.md
+++ b/charts/rstudio/CHANGELOG.md
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+## [1.5.3] - 2018-06-05
+### Changed
+- Adjusted memory resource limits and requests from `request: 1GB => 5GB` & `limit: 12GB => 20GB`
+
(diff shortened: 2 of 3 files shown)


charts/rstudio/values.yaml
```diff
@@ -35,7 +35,7 @@ rstudio:
   resources:
     limits:
       cpu: 1.5
-      memory: 12Gi
+      memory: 20Gi
     requests:
       cpu: 200m
-      memory: 1Gi
+      memory: 5Gi
```
xoen: You probably discussed this with Robin and you know this better than me, but increasing the requested memory that means all RStudio instances will want 5GB minimum even if they'll not use them, and also the worker nodes will be able to host fewer RStudio instances because of this higher requirements. Can't remember how much memory each node has.

I'm not saying it's bad, as long as we're happy with this :) 

(If I remember well how requests/limits params work in k8s!)

djsd123: That's correct.  The thinking behind this is there will be less interruption to the user's since the scheduler has a tendency to bin pack.  So if a user tries to burst its limit the pod/pods get rescheduled. We have `r4.xlarge` = `30GB`. We'll see how our instances hold up but I think another IG specifically for rstudio is what I would do. Node Affinity.

xoen: 👍

djsd123: Indeed.  Usage, from what I have seen is sporadic. If anything I am more worried about the SoftNAS. Now that we have a `20GB` limit, the core dumps potentially become larger in case of a crash. I am not too sure how rstudio handles crashes or whether it cleans up after itself.  I was considering extending the termination grace period but I want to see what effect this has first.  We're deploying this to early adopters first.

andyhd:
@djsd123 this is still open - is it ok to merge now?

djsd123:
Prepare for carnage 💯

PR #181, opened 2018-06-18
Helm: Upgrade Webapp script

djsd123:
Interactive shell script to perform chart upgrades for shiny-apps in k8s namespace: `apps-prod`

We use `helm get values` to ensure we have all values required to perform
an upgrade. This does pose questions about conflicts during a user
provided values merge and ideally we want whatever is in the repo to take precedence
but for now, this solution provides the most predictable behaviour.    
Please see issue [here](https://github.com/kubernetes/helm/issues/3957) for more information

__With__ or __Without__ flags
- Select `y` when prompted to perform __dry_run__

With defaults:
1. Run script (./upgrade_release.sh) with no flags

With Options:

1. Run script (./upgrade_release.sh) with `-h` to view usage

```
   usage: All arguments are optional
  -e - Environment e.g. dev or alpha, Default: alpha
  -d - Debug: true or false, Default: false
  -f - File: Path to config/values file, i.e. analytics-platform-config/chart-env-config/<ENV>/webapp.yml
  -n - Namespace: Kubernetes namespace, Default: apps-prod
  -s - Save release values yaml files: true or false, Default: false
  -h - Print usage

```
2. If you wish? Test each flag
3. Run script with an invalid option like `-y`.  This should result in
usage

### DIFF
--- a/charts/webapp/migrations/upgrade_releases.sh
+++ b/charts/webapp/migrations/upgrade_releases.sh
@@ -1,0 +1,89 @@
+#!/usr/bin/env bash
+
+set -uo pipefail
+
+DEBUG=false
+DRY_RUN=true
+CONFIG_PATH=
+ENV=alpha
+HELM_REPO=mojanalytics/webapp
+HELM_VER=2.9.1
+NAMESPACE=apps-prod
+MOJANALYTICS_REPO=http://moj-analytics-helm-repo.s3-website-eu-west-1.amazonaws.com
+PLATFORM=$(uname -s)
+SAVE_VALUES=false
+
+
+usage() {
+  cat << EOF
+  usage: All arguments are optional
+  -e - Environment e.g. dev or alpha, Default: alpha
+  -d - Debug: true or false, Default: false
+  -f - File: Path to config/values file, i.e. analytics-platform-config/chart-env-config/<ENV>/webapp.yml
+  -n - Namespace: Kubernetes namespace, Default: apps-prod
+  -s - Save release values yaml files: true or false, Default: false
+  -h - Print usage
+EOF
+}
+
+while getopts e:d:f:n:h option; do
+  case $option in
+    e   ) ENV="$OPTARG";;
+    d   ) DEBUG="$OPTARG";;
+    f   ) CONFIG_PATH="$OPTARG";;
+    n   ) NAMESPACE="$OPTARG";;
+    s   ) SAVE_VALUES="$OPTARG";;
+    h   ) usage; exit 0;;
+    \?  ) usage; exit 1;;
+  esac
+done
+
+# If CONFIG_PATH has a value prefix it with "--values"
+    if [[ ! -z $CONFIG_PATH && -f $CONFIG_PATH ]]; then
+
+      CONFIG_PATH="--values ${CONFIG_PATH}"
+
+    elif [[ ! -z $CONFIG_PATH && ! -f $CONFIG_PATH ]]; then
+
+      Printf "\nCannot find config file: ${CONFIG_PATH}.  Are you sure it exists or is a file?\n"
+      exit 1
+
+    fi
+
+while true; do
+	read -p "Is this a DRY RUN? i.e. NO-OP [y/N]: " dry_run
+	case $dry_run in
+	    [y]*    ) DRY_RUN=true; break;;
+	    [N]*    ) DRY_RUN=false; break;;
+	    *       ) echo "Answer y to simulate the operation.  Answer N to perform the operation."; exit;;
+    esac
+done
+
+# Grab the Helm client if not installed
+which helm || wget --progress=dot:giga -O helm https://storage.googleapis.com/kubernetes-helm/helm-v$HELM_VER-$PLATFORM-amd64.tar.gz
+
+if [ -f helm ]; then
+  PATH="$PATH:$(pwd)"
+  chmod +x helm
+fi
+
+# Ensure we've added our helm repo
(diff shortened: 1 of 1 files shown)


djsd123:
@xoen, will this cause that issue where ci goes nutz and does loads of commits?

xoen:
@mikekeen1 it shouldn't, we don't re-build the Helm repo index using CircleCI anymore, Concourse does that and it doesn't commit on this repository (as repository is in S3 now).

I only had a look at the README. Regarding option, I think I'd err on the side of caution and have `-e` (environment) mandatory to be explicit. Having a look at the code now.

charts/webapp/migrations/upgrade_release.sh
```diff
@@ -0,0 +1,66 @@
...
+    d   ) DEBUG="$OPTARG";;
+    f   ) CONFIG_PATH="$OPTARG";;
+    n   ) NAMESPACE="$OPTARG";;
+    h   ) usage; exit 0;;
+    \?  ) usage; exit 1;;
+  esac
+done
+
+if [ ! -f $CONFIG_PATH ]; then
+  Printf "\nCannot find config file: ${CONFIG_PATH}.  Are you it exists or is a file?\n"
```
xoen: 👍 for helpful message

(there is a small typo in the question, I think it misses "sure", "Are you sure it exists and it's a file?")

charts/webapp/migrations/upgrade_release.sh
```diff
@@ -0,0 +1,66 @@
...
+  exit 1
+fi
+
+while true; do
+	read -p "Is this a DRY RUN? i.e. NO-OP [y/N]: " dry_run
+	case $dry_run in
+	    [y]*    ) DRY_RUN=true; break;;
+	    [N]*    ) DRY_RUN=false; break;;
+	    *       ) echo "Answer y to simulate the operation.  Answer N to perform the operation."; exit;;
+    esac
```
xoen: `esac` indentation is a bit off, not that this causes problems.

charts/webapp/migrations/upgrade_release.sh
```diff
@@ -0,0 +1,66 @@
...
+# Grab the Helm client if not installed
+which helm || wget --progress=dot:giga -O helm https://storage.googleapis.com/kubernetes-helm/helm-v$HELM_VER-$PLATFORM-amd64.tar.gz
+
+if [ -f helm ]; then
+  PATH="$PATH:$(pwd)"
+  chmod +x helm
+fi
+
+# Ensure we've added our helm repo
+helm repo list | grep -F 'mojanalytics' && helm repo update || helm repo add mojanalytics http://moj-analytics-helm-repo.s3-website-eu-west-1.amazonaws.com
```
xoen: Parenthesis here could help for the slow people like me :), e.g.

```bash
(helm repo list | grep -F 'mojanalytics' && helm repo update) || helm repo add mojanalytics http://moj-analytics-helm-repo.s3-website-eu-west-1.amazonaws.com
```

Does `helm repo add` also does the update right? I don't remember. If it does, this is fine otherwise you may have to to the `repo update` manually.

I would probably add the mojanalytics repo URL in a constant as well.

charts/webapp/migrations/upgrade_release.sh
```diff
@@ -0,0 +1,66 @@
...
+
+for shiny_app in $SHINY_APPS; do
```
xoen: For some reason have the command to get the shiny apps in a variable at the top of the file feels off. I'm not sure I'd use a constant/variable in this case (but maybe I'm missing something) and if I did, I'd probably keep it near the place where it's used. IMHO of course :)

charts/webapp/migrations/upgrade_release.sh
```diff
@@ -0,0 +1,66 @@
...
+  helm upgrade --reuse-values=true --recreate-pods=true --dry-run=$DRY_RUN --namespace=$NAMESPACE $shiny_app $HELM_REPO \
```
xoen: Uh I didn't know there was this `--recreate-pods` flat available, that's useful! Especially when configmaps change, environment variables, etc, in the pods don't change. Nice 👍 

Is the `--reuse-values` doing anything if we're re-passing in the config yaml?

charts/webapp/migrations/upgrade_release.sh
```diff
@@ -0,0 +1,66 @@
...
+    --debug=$DEBUG -f $CONFIG_PATH
+  sleep 3
```
xoen: Why do we `sleep` here?

mikekeen1:
Hi,
Why am I getting these emails? Are you able to take me off the list please.
Mike

charts/webapp/migrations/upgrade_release.sh
djsd123: Because it's an interactive script, it makes for a more palatable output but serves no real function

charts/webapp/migrations/upgrade_release.sh
djsd123: Must be a github thing.  Looks right in my editor.  I'll retab to see if that helps

charts/webapp/migrations/upgrade_release.sh
djsd123: Refer to our discussion 👎

djsd123:
My take on the whole thing is maybe we're using helm incorrectly and that user-supplied values or config in our case should represent everything you don't hard code in the template.  So for everything you define in a chart's `values.yaml` you should also have it present in config.

PR #183, opened 2018-06-26
Logstash Chart

djsd123:
Deploys logstash with preconfigured http input, output to elasticsearch and x-pack
monitoring.

### DIFF
--- a/charts/logstash/templates/service-account.yaml
+++ b/charts/logstash/templates/service-account.yaml
@@ -1,0 +1,9 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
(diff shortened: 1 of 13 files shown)


charts/logstash/templates/service-account.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ServiceAccount
```
kerin: Do we need a ServiceAccount if there's no Role associated with it?

r4vi: I prefer this way as you get a named service account instead of one called `default`

PR #186, opened 2018-07-09
Upgrade Logstash

djsd123:
Changing image tag to match elasticsearch version

### DIFF
--- a/charts/logstash/CHANGELOG.md
+++ b/charts/logstash/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [0.1.1] - 2018-07-09
+## Upgrade Logstash
+- Bumping image tag to 6.3.1 (from 5.6.10) to match elasticsearch version
+
 ## [0.1.0] - 2018-06-26
 ### Initial Commit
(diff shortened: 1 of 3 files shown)


charts/logstash/values.yaml
```diff
@@ -3,7 +3,7 @@ replicaCount: 1
 
 image:
   repository: docker.elastic.co/logstash/logstash
-  tag: 5.6.10
+  tag: 6.3.1
```
xoen: Any breaking change in this?

charts/logstash/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [0.1.1] - 2018-07-09
+## Upgrade Logstash
+- Modifying Image tag to match elasticsearch version
```
xoen: I guess it's overkill as we can always go back, check Git history, etc...but, I find useful to have version changes in the Changelog as well, e.g.

> Bumping image tag to `6.3.1` (from `5.6.10`) to match elasticsearch version

charts/logstash/values.yaml
djsd123: Breaks without it

charts/logstash/CHANGELOG.md
djsd123: 👍

PR #188, opened 2018-07-10
Airflow configured to run tasks on new pods (KubernetesExecutor)

xoen:
It's mostly a usual helm chart. There are few bits which are required to make
the mounting of the DAG/logs volumes work, etc...

### Airflow + Kubernetes configuration
The main feature of this helm chart is the fact that Airflow is configured to
use the `KubernetesExecutor`. In contrast to the traditional way of setting up
Airflow, where some workers are sitting there waiting for some task to be
scheduled to them, tasks will be run in a newly created kubernetes pod.

### DAGs location
There are more details on where DAGs will be located below but it's important to
understand that:
1. The authoritative source of true regarding the DAGs will be a [Git
   repository](https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/188/files#diff-d6074f20d25632dde0b1c254089d138cR35)
2. The DAGs will be periodically synced from this repository into a persisted
   volume (every [3m](https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/188/files#diff-d6074f20d25632dde0b1c254089d138cR34)
   by default)

This is convenient and has other benefits:
- newly created pods don't have to pull the repository every time
- all the Airflow components see pretty much the same version of the DAGs
  because they all mount the same volumes (see below)

### Airflow image
We're building/using our own Airflow docker image. This is done because the
development of airflow and specifically the new shiny k8s things is relatively
fast.

Essentially we do this to have control over the "version" of airflow used and to
not have flaky containers.

- [Dockerfile is here](https://github.com/ministryofjustice/analytics-platform-airflow-docker-image)
- Image is built on quay.io when a GH tag is created
- Image is hosted in [quay.io](https://quay.io/repository/mojanalytics/airflow?tab=tags): `quay.io/mojanalytics/airflow`

The Dockerfile pins the version of Airflow, currently it's on a git sha of a
well known good version of Airflow. Once v1.10 is released we probably wont need
to pip install from a sha and can switch to installing it from pypi directly.

The Dockerfile is still required to add the
[fab-oidc](https://pypi.org/project/fab-oidc/) dependency that is used for
logging in using Auth0.

#### PVs (PersistedVolumes)
The helm chart uses two PVc (PersistedVolumes) which are backed by a EFS volume
(in our case). The EFS volume is [created via
terraform](https://github.com/ministryofjustice/analytics-platform-ops/pull/148)

This volume is used for two reasons:
- To store the DAGs and share these across the webserver, scheduler and tasks
  containers
- To keep the logs. As the tasks are run in pods/containers, their logs will be
  lost once they're terminated. And that's why we keep them in a PV.

#### PVC (PersistedVolumeClaim)
There are also two PVCs for the corresponding DAGs and logs PVs.

#### Reasons for having two PVC
Mainly it's because the `KubenetesExecutor` has two config for the DAGs and the
logs PVC (`kubernetes.dags_volume_claim` and `kubernetes.logs_volume_claim`).

We also had to create two PV for few reasons:
1. Different PVCs can't claim the same PV in the same namespace (main)
2. PVC can't specify a path anyway
3. If you had 1 PVC + 1 PV and mounted twice in the same Pod with different
   subPaths then Kubernetes
   fails to mount them. We think this is a [bug in Kubernetes](https://groups.google.com/forum/#!topic/openshift/klwlwl2m-uE)

#### Initialisation of the DAGs/logs PVCs
The PVs have a `nfs.path` attribute of `/dags` and `/logs` respectively. When
the EFS volume is empty these directories don't exist. This will cause problems
the first time the corresponding PVC are mounted.

For this reason there is a job (`{{ .Release.Name }}-create-nfs-dirs`) which
creates the `/dags` and `/logs` directory on the EFS volume (calling `mkdir -p`
so it's idempotent).

#### Getting rid of NFS
In general all of this adds complexity, and we might consider moving the logs to
S3 and the DAGs to sync with git in every pod. We haven't done this yet as
configuring logging is changing in Airflow and it's not well documented or
stable yet.

### DAGs syncing
There is a pod which is responsible to pull and periodically sync the DAGs from
a Git repository into the DAGs PVC.

The chart values under `gitSync` can be used to control:
- the `repository` where the DAGs are (this currently needs to be public),
  defaults to [`"https://github.com/ministryofjustice/analytics-platform-airflow-example-dags"`](https://github.com/ministryofjustice/analytics-platform-airflow-example-dags)
- the frequency of syncing by changing the `interval` value. Default is `180`
  seconds (3 minutes)
- the `branch` to sync. Defaults to `master`

A note on syncing: Here I'm talking about syncing but it's not a two-way
syncing, it's just unidirectional from the Git repository to the DAGs volume.
Nothing is pushed to the Git repository.

### Database
Airflow is configured to use a Postgres database. In our case this is an RDS
instance hosted in AWS. This is created [via
terraform](https://github.com/ministryofjustice/analytics-platform-ops/pull/148).

### ServiceAccounts
There are two service accounts: `{{ .Release.Name }}` (probably `airflow` most
of the time) and `{{ .Release.Name }}-tasks` (e.g. `airflow-tasks`).

The first is the one used by the webserver and scheduler. The `*-tasks` service
account is used by the tasks' pods created when running a pod.


charts/airflow-k8s/templates/deployment.yml
```diff
@@ -0,0 +1,116 @@
...
+      volumes:
+        - name: {{ .Release.Name }}-config
+          configMap:
+            name: {{ .Release.Name }}
+        - name: {{ .Release.Name }}-dags
+          persistentVolumeClaim:
+            claimName: {{ .Release.Name }}-dags
+        - name: {{ .Release.Name }}-logs
+          persistentVolumeClaim:
+            claimName: {{ .Release.Name }}-logs
```
r4vi: new line at end of file

charts/airflow-k8s/templates/service-account.yaml
```diff
@@ -0,0 +1,142 @@
+apiVersion: v1
```
r4vi: maybe rename this to `.yml` instead of `.yaml` for consistency?

charts/airflow-k8s/templates/deployment.yml
xoen: Weird, not sure how that happened. I'll add 👍

charts/airflow-k8s/templates/service-account.yaml
xoen: Good point 👍

(I wonder why we don't just stick with `.yaml` for everything, what is this? DOS 😆 )

xoen:
@r4vi I also [updated the README](https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/188/commits/8c589c7ec11409fad0ad34f221441abe3bd6255e) to document some of the other values, have a look and see if that makes sense, etc...

xoen:
Both: let me know if this can be merged as it is or there is any blocker.

r4vi:
@xoen LGTM :+1:

xoen:
@djsd123 also happy with this?

r4vi:
@djsd123 @xoen will it automatically be deployed by CI after merge or will I have to do anything?

djsd123:
Sorry, been desperately trying to fix elasticsearch.  Will look at this next if you still need my input?

charts/airflow-k8s/templates/service-account.yml
```diff
@@ -0,0 +1,142 @@
...
+subjects:
+  - kind: ServiceAccount
+    name: {{ .Release.Name }}
+    namespace: {{ .Release.Namespace }}
+roleRef:
+  kind: Role
+  name: pod-logs
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
```
djsd123: No harm but any reason why you're using `v1beta1` here?  `Rolebinding` should be available in the stable release of the API like above ^^.  Also, may avoid future deprecations..

djsd123: Yep, but only if you can be bothered..  It'll work as is.  I'd just merge tbh

xoen: Thanks for the review @djsd123 - I'm torned between merging a working version of this chart or change this before merging.

The right thing to do is to change tho, so I'll try to do an `helm delete --purge` in `dev` and try to re-install and see if it complains, if not I'll change.

xoen: FYI: Tested in `dev` and it seemed to work. [Updated and pushed](
https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/188/commits/32b468b66a15e77bc19343b29d54a51003375b1)a. Merging this 🎊.

PR #194, opened 2018-07-19
Airflow: decouple webserver and scheduler

r4vi:
This decouples the webserver from the scheduler pod. Allowing us to scale the
webserver independently from the scheduler. This is important because airflow
can only having one running scheduler connected to a single db.


r4vi:
related to https://github.com/ministryofjustice/analytics-platform-config/pull/71

charts/airflow-k8s/templates/scheduler-deployment.yml
```diff
@@ -37,7 +40,7 @@ spec:
           - "-cx"
           - >
               cd /usr/local/lib/python3.6/site-packages/airflow &&
-              airflow initdb && airflow upgradedb &&
+               airflow initdb && airflow upgradedb &&
```
r4vi: oops extra space

charts/airflow-k8s/templates/scheduler-deployment.yml
```diff
@@ -1,22 +1,25 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: {{ .Release.Name }}
+  name: {{ .Release.Name }}-scheduler
```
djsd123: Shouldn't we be namespacing these?

r4vi: like: `{{ .Release.Name }}-{{ .Release.Namespace}}-scheduler`?

r4vi:
@djsd123 what do you mean by namespacing it? just this? https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/194#discussion_r203734220

charts/airflow-k8s/templates/scheduler-deployment.yml
djsd123: Meant explicitly specifying namespaces in metadata like:

```
metadata:
  name: {{ .Release.Name }}-webserver
  namespace: {{ .Release.Namespace }}
  labels:
    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
```

r4vi:
@djsd123 I've added namespaces everywhere we forgot, even PVs (because it's useful to know which namespace they're used in) - is this ok or did I go overboard?

charts/airflow-k8s/templates/scheduler-deployment.yml
xoen: Isn't that done by helm when you specify the namespace argument?

xoen:
@r4vi / @djsd123 I thought PV (persistent volumes) were not living in a specific namespace, but PVC were (persistent volume claims) 🤷‍♂️

charts/airflow-k8s/Chart.yaml
```diff
@@ -1,4 +1,5 @@
 apiVersion: v1
 description: Airflow on kubernetes. Tasks are run as pods
 name: airflow-k8s
-version: 0.0.1
+icon: https://airflow.incubator.apache.org/_images/pin_large.png
```
xoen: ✨

charts/airflow-k8s/templates/volumes.yml
```diff
@@ -20,6 +20,7 @@ kind: PersistentVolume
 apiVersion: v1
 metadata:
   name: {{ .Release.Name }}-dags
+  namespace: {{ .Release.Namespace }}
```
xoen: As mentioned, I thought PVs lived outside specific namespaces unlike PVC which sits within a specific namespace.

Also, don't we need to specify the `Kind` for this PV resource? I've just noticed this.

r4vi:
@xoen I just tagged the PVs with namespace so we can see what namespace they are for, even if they are not namespaced objects

charts/airflow-k8s/Chart.yaml
r4vi: it was to stop `helm lint` complaining :1st_place_medal:

charts/airflow-k8s/templates/volumes.yml
r4vi: kind is on line 19 (github collapsing it)

xoen: Uh, I expanded it but missed, damn GH 👍

xoen:
@r4vi oh so you can add a namespace to PV even if k8s ignores them? I wonder if it's more of a label job.
Maybe this is a common pattern I'm not aware of 👍

r4vi:
@xoen it's not a pattern I know of, we're all making this up as we go along :smile: happy to get rid of it

r4vi:
@xoen I deployed the chart and checked the PV that k8s has and it seems to ignore it totally so I'll remove it

the result:
```yaml
apiVersion: v1
kind: PersistentVolume
metadata:
  annotations:
    pv.kubernetes.io/bound-by-controller: "yes"
    volume.beta.kubernetes.io/storage-class: nfs
  creationTimestamp: 2018-07-17T17:30:48Z
  labels:
    app: airflow
    chart: airflow-k8s-0.0.2
    name: airflow-dags
  name: airflow-dags
  resourceVersion: "100850421"
  selfLink: /api/v1/persistentvolumes/airflow-dags
  uid: 24c9a457-89e7-11e8-81f8-02b0f61ecc64
spec:
  accessModes:
  - ReadWriteMany
  capacity:
    storage: 10Gi
  claimRef:
    apiVersion: v1
    kind: PersistentVolumeClaim
    name: airflow-dags
    namespace: airflow
    resourceVersion: "100259211"
    uid: 24d0672d-89e7-11e8-81f8-02b0f61ecc64
  nfs:
    path: /dags
    server: fs-36e588ff.efs.eu-west-1.amazonaws.com
  persistentVolumeReclaimPolicy: Retain
status:
  phase: Bound

```

charts/airflow-k8s/templates/scheduler-deployment.yml
djsd123: Yes, but namespacing explicitly is good practice because it ties things together like where your pods will be deployed and things like serviceaccounts and `RBAC` can be expressed and glued together properly.

PR #195, opened 2018-07-23
Helm: RBAC Chart

djsd123:
Initial commit

A chart to deploy miscellaneous `rbac` resources for users (mainly)

[Trello](https://trello.com/c/ZK7eHGqB)

### DIFF
--- a/charts/rbac/README.md
+++ b/charts/rbac/README.md
@@ -1,0 +1,44 @@
+# RBAC Resources
+
+Chart to deploy user and group [RBAC](https://kubernetes.io/docs/reference/access-authn-authz/rbac/) resources
+
+### Installing and Upgrading a release
+
+```
+helm upgrade rbac-chart analytics-platform-helm-charts/charts/rbac --install
+```
+
+## Configuration
+
+| Parameter  | Description      | Default |
+| ---------- | ---------------  | ------- |
+| `team`     | Github team name as it appears in github |   ""  |
+| `namespace` | Kubernetes namespace for namespaced scoped resources i.e. `Role`, `Rolebinding` | "" |
+| `description` | Short description of RBAC resources use | "" |
+
+
+## Contributing 
+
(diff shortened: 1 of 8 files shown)


charts/rbac/README.md
```diff
@@ -0,0 +1,9 @@
+# RBAC Resources
+
+Chart to deploy miscellaneous RBAC resources
+
+### Installing and Upgrading a release
+
+```
+helm upgrade rbac-chart analytics-platform-helm-charts/charts/rbac --install
+```
```
andyhd: This readme is a bit terse. What are the miscellaneous RBAC resources? Why might I want to install this chart?

charts/rbac/templates/_helpers.tpl
```diff
@@ -0,0 +1,13 @@
+{{/* vim: set filetype=mustache: */}}
```
andyhd: Even though I am a die-hard vim fan, I think you should keep your settings in your `.vimrc` file and not have them here.

andyhd: Not to mention that this is not a mustache file.

charts/rbac/templates/_helpers.tpl
```diff
@@ -0,0 +1,13 @@
...
+Expand the name of the chart.
+*/}}
+{{- define "rbac.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
+Create chart name and version as used by the chart label.
+*/}}
+{{- define "rbac.chart" -}}
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" -}}
```
andyhd: Is this algorithm a standard? If the chart name is too long, the version will be truncated first. That seems wrong.

djsd123:
@andyhd @davidread @r4vi  ^^

PR #205, opened 2018-08-13
Cert-Manager Resources

djsd123:
[Trello](https://trello.com/c/M1snktNZ)

Initial commit

- Provisions a [cluster-issuer](https://cert-manager.readthedocs.io/en/latest/reference/clusterissuers.html) resource
  - This resource defines the request to the CA and any validation challenges   

- Provisions a [certificate](https://cert-manager.readthedocs.io/en/latest/reference/certificates.html) resource
  - This resource defines the tls cert 

Relates to [PR](https://github.com/ministryofjustice/analytics-platform-config/pull/76)

### DIFF
--- a/charts/cert-manager-resources/templates/certificate.yaml
+++ b/charts/cert-manager-resources/templates/certificate.yaml
@@ -1,0 +1,22 @@
+apiVersion: certmanager.k8s.io/v1alpha1
+kind: Certificate
+metadata:
+  name: {{ .Release.Name }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: cert-manager
+spec:
+  # Name of a secret to store tls certificate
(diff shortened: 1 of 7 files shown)


charts/cert-manager-resources/templates/certificate.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: certmanager.k8s.io/v1alpha1
+kind: Certificate
```
xoen: So this API and resource are added/handled by the cert-manager (installed via other helm chart) right?

djsd123: yep.  Cert-Manager Operator

r4vi:
I'm merging this so we have the currently deployed version of this committed then I can iterate on it

PR #209, opened 2018-09-03
Improvements following manual set-up of a platform

davidread:
Includes:
* What charts are needed for a 'minimal set-up' of the platform
* helm init needs the `--service-account` param
* Auth0 application config (rstudio, cpanel)
* Add info on important charts: kube2iam, rbac


charts/README.md
```diff
@@ -27,9 +27,43 @@ by running these commands from the root of the project (change `dev` to your env
 Necessary to access the services from outside the cluster.
 
 ```bash
-$ helm upgrade cluster-ingress charts/nginx-ingress -f chart-env-config/<ENV>/nginx-ingress.yml --namespace kube-system --install
+$ helm upgrade cluster-ingress charts/nginx-ingress -f ../analytics-platform-config/chart-env-config/dev/nginx-ingress.yml --namespace default --install
 ```
 
+### Domain SSL certificate
```
davidread: I've considered moving the first part of this to the ops repo's README, but since it's so closely coupled with configuring the chart, it feels better here as a whole.

charts/cpanel/README.md
```diff
@@ -23,35 +23,50 @@ The instance will be available at https://cpanelapi-$BRANCH_NAME.<ServicesDomain
 
 ## Configuration
 
+### Auth0 application
```
davidread: I'd like to write terraform for this, but in the meantime is it better for these Auth0 instructions to live here, rather than -ops?

davidread:
@kerin I've made a couple of comments/questions in this PR - what do you think on those? (Don't worry about reviewing the detail of the PR - no doubt someone on our team can do that)

charts/README.md
kerin: @davidread thanks to the magic of not looking at this for a month, we don't need this section at all now, as we've moved to nginx-ingress + NLB + Let's Encrypt, so there's no need for manual SSL setup anymore

charts/cpanel/README.md
kerin: @davidread yep makes sense - probably want to link to this from `analytics-platform-ops` readme though, as chart docs are easy to miss, and without doing this authentication for RStudio etc isn't going to work

davidread: Great, I've added a note: https://github.com/ministryofjustice/analytics-platform-ops/pull/191/files#diff-04c6e90faac2675aa89e2176d2eec7d8R513

charts/README.md
davidread: Hooray for simplification!!

charts/cpfrontend/values.yaml
```diff
@@ -3,7 +3,7 @@ Frontend:
   Branch: "master"
   Image:
     Repository: quay.io/mojanalytics/control-panel-frontend
-    Tag: 0.1.0
+    Tag: v1.3.14
```
davidread: 0.1.0 isn't a release any more

README.md
```diff
@@ -46,7 +46,7 @@ To run a minimal version of AP, these are the essential charts to deploy:
 4. [Test in dev](#how-to-test)
 5. Push the changes
 
-Commits to the master branch of this repository will trigger a Concourse CI pipeline which updates the packaged Helm charts and index in our Helm repository:
+Commits to the master branch of this repository will trigger a [Concourse CI pipeline](https://github.com/ministryofjustice/analytics-platform-concourse-pipelines#update-helm-repoyaml) which updates the packaged Helm charts and index in our Helm repository:
```
r4vi: typo? repo.yaml?

davidread: not a typo - I guess dots are stripped by github's markdown processor to make the fragment name look beautiful

davidread:
This is updated. Is it ok to merge @ravi?

r4vi:
@davidread LGTM, btw I'm @r4vi not @ravi on here

davidread:
oops! sorry about the name!

PR #210, opened 2018-09-03
Traefik compatibility

davidread:
This is recording what I used successfully on dev today, to work with the Traefik ingress, currently running there.

I guess we'd merge this only at the point of deploying Traefik to alpha. In the meantime, we just remember what we need on dev?


r4vi:
I think we can close this since we're not using træfik anymore

r4vi:
actually I've just tried istio and I think removing the ingress annotation completely works too (and allows switching between ingress much more easily)

davidread:
Happy to have an isteo-specific version of this :)

PR #211, opened 2018-09-03
Concourse: Target correct pipeline

r4vi:
The old chart used to target a pipeline called `moj-analytical-services` because
the pipeline name was based on the org name. At some point we manually deployed
this to a pipeline called `create-webapp-pipelines` in the admin team.

These changes update the chart to reflect the current state of the world.

I've also updated the pipeline to read the github tokens from k8s secrets
instead of having them passed in with fly

### DIFF
--- a/charts/concourse-org-pipeline/templates/job.yaml
+++ b/charts/concourse-org-pipeline/templates/job.yaml
@@ -5,6 +5,10 @@ metadata:
   namespace: concourse-{{ .Values.concourse.team }}
   labels:
     chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+  annotations:
+    "helm.sh/hook": post-upgrade
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": hook-succeeded
 spec:
   template:
     metadata:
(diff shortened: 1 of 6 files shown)


r4vi:
added `don't merge` label because the config will need to be updated to include the username/password for the `admin` team not the `main` team

charts/concourse-org-pipeline/templates/job.yaml
```diff
@@ -15,20 +19,19 @@ spec:
...
           - name: CONCOURSE_TEAM
             value: {{ .Values.concourse.team }}
+          - name: CONCOURSE_TARGET_TEAM
+            value: {{ .Values.concourse.target_team }}
           - name: CONCOURSE_URL
             value: {{ .Values.concourse.externalURL }}
           - name: FLY_PASSWORD
-            valueFrom:
-              secretKeyRef:
-                name: fly
```
r4vi: this was the target team password so we can't use it from the ((fly.secrets))

r4vi:
@xoen I've made fly username and password come from secrets now

r4vi:
[Update update-helm-repo.yaml pipeline to use secrets](https://trello.com/c/PxP26Vp4/1332-update-update-helm-repoyaml-pipeline-to-use-secrets)

PR #213, opened 2018-09-11
Add job for the controlpanel migration

davidread:
Because the migration is now not done in Dockerfile startup. See: https://github.com/ministryofjustice/analytics-platform-control-panel/pull/335

### DIFF
--- a/charts/cpanel/templates/job.yaml
+++ b/charts/cpanel/templates/job.yaml
@@ -1,0 +1,33 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ template "fullname" . }}-migration
+spec:
+  template:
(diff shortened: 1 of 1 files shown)


davidread:
I have retested this on the accelerator cluster:
```
helm delete cpanel-master
helm upgrade cpanel-master charts/cpanel -f ~/gds/data-science-accelerator-config/chart-env-config/accelerator/cpanel.yml --set API.Image.Repository=quay.io/david_read/analytics-platform-control-panel --set API.Image.Tag=0.5
```
and it works. It takes 10s or more to return from the `helm upgrade` but the job runs fine and is successfully deleted, allowing `helm upgrade` to be rerun successfully.

PR #214, opened 2018-09-12
airflow-k8s 0.0.5: Fixed missing SMTP credentials in k8s operator pod

xoen:
This was causing emails to not be sent.


xoen:
Thanks @r4vi 🏆

PR #215, opened 2018-09-13
airflow-k8s 0.0.6: Fixed URLs in email notifications

xoen:
URLs in emails were previously `http://localhost:80` which is obviously nonsense.
This fixes it.


xoen:
Cheers @r4vi 🌟

PR #216, opened 2018-09-20
Istio: Added README with basic installation instructions

xoen:
**NOTE**: This is not an helm chart we maintain, we're using the official
Istio helm chart. Here I just added some notes in case someone looks
for Istio in this repository.


davidread:
If we're not checking the chart into here, then I kind of think we should define our current isteo release/chart version somewhere. If not here, then where? We seem to have a gap in our docs this repo and the [ops readme](https://github.com/ministryofjustice/analytics-platform-ops) - nowhere says "now install these particular charts in this order".

charts/istio/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+
+### Installation
+We're using the official Istio helm chart to install Istio in our k8s cluster.
+The Istio website has installation instructions [here](https://istio.io/docs/setup/kubernetes/helm-install/),
+please refer to Istio website for up-to-date instructions.
+
+
+#### Download Istio release
+First, [download the release](https://istio.io/docs/setup/kubernetes/download-release/):
```
davidread: I suggest you say that the release contains the helm chart.

charts/istio/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+#### Install custom resource definitions (CRDs) if necessary
+
+- Check helm version `$ helm version`
+- If version is prior to `2.10.0` you'll also need to install custom resource
+  definitions (CRDs) manually as follow:
+  `$ kubectl apply -f install/kubernetes/helm/istio/templates/crds.yaml`
+
+
+### Install helm chart
```
davidread: I think this should be `####` so that it is in the 'install' section

charts/istio/README.md
```diff
@@ -0,0 +1,63 @@
...
+Configuration for the helm chart is in the usual [config repository](https://github.com/ministryofjustice/analytics-platform-config)
+
+Specifically you'll have to pass the values when installing the helm chart:
+
+```bash
+$ helm install install/kubernetes/helm/istio --name istio --namespace istio-system -f $PATH_TO_CONFIG/chart-env-config/$ENV/istio.yml
+```
+
+
+### Uninstall helm chart
```
davidread: Just 'Uninstall' is probably more accurate, seeing as you have the bit about CRDs.

charts/istio/README.md
```diff
@@ -0,0 +1,63 @@
+## Istio
+
+
+### What is Istio
+[Istio](https://istio.io) is an open platform-independent service mesh that provides traffic
+management, policy enforcement, and telemetry collection.
+
+Istio has many components and features but initially we're only using it to
+replace nginx-ingress.
```
davidread: Can you describe it without having to refer to nginx-ingress, which will be gone from this project shortly? i.e. say what we are using isteo for.

charts/istio/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+
+### Installation
+We're using the official Istio helm chart to install Istio in our k8s cluster.
```
davidread: Maybe it will be obvious when this PR is merged, but I think it could be clearer to say that this repo doesn't include a helm chart. And "Instead download the Istio release and use the helm chart in that".

charts/istio/README.md
xoen: I'll replace with:

> Istio has many components and features but initially we're only using it to
direct traffic from the internet into right pod (Ingress resources).

charts/istio/README.md
xoen: Good point, I'll change this section to:

> This repository does *not* include the helm chart for Istio - we're using the
official helm chart which is part of the Istio release (see below).

> This repository includes instructions on how to install Istio in kubernetes
but for up-to-date documentation refer to the [Istio's webside](https://istio.io/docs/setup/kubernetes/helm-install/).

xoen:
@davidread I didn't put much emphasis on chart version as I suspect they'll update it regularly so it may be best to try to stay up-to-date. Bear in mind you can see version of Istio installed using `$ helm list istio` and also we may decide to pin the version of docker image used in config/values.

Basically I tried to keep this small and version agnostic.

charts/istio/README.md
xoen: Well spotted, I clearly missed an hash 👍

charts/istio/README.md
xoen: Good point, I'm changing to:

> Istio keeps the official helm chart inside its release.

charts/istio/README.md
xoen: Good point. I must have added the `$ kubectl delete` command later on, you're right I'll change to `Uninstall` 👍

xoen:
FYI: https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/216/commits/eee07d9d3a5610599a3a032771a4b8e47ebf6d05

PR #217, opened 2018-09-24
Cert Manager Resources: update for Istio

r4vi:


r4vi:
guess we don't need this anymore

PR #219, opened 2018-09-26
Added tls block in ingress resources

xoen:
This is needed by nginx-ingress. This is added by default but
you can set the `ingress.addTlsBlock` value to `false`.

- [x] kibana-auth-proxy
  - release: `cluster-logviewer`
  - ns: `kube-system`
- [x] kubernetes-dashboard ? 
  - ns: `kube-system`
  - NOTE: This is in our repo. We do **NOT** use the [stable/kube-dashboard](https://github.com/helm/charts/tree/master/stable/kubernetes-dashboard) chart. We **should** move to it at some point
- [x] release: `kube-prometheus`
  - ns: `monitoring`
  - **NOTE**: I *think* this is the helm chart that was installed: [coreos/kube-prometheus](https://github.com/coreos/prometheus-operator/tree/master/helm/kube-prometheus)
  - **NOTE**: Stick with the same version (`0.0.32`) for now as newer version changed values structures and even when I tried to fix them I got some other errors which I think may be related to CRDs, `helm upgrade --dry-run --debug kube-prometheus coreos/kube-prometheus --version 0.0.32` seemed promising.
- [ ] Grafana
  - release: `cluster-monitoring`
  - Release in `alpha` is using `grafana-0.8.0` but user-supplied-value for `server.image` is a forked image [`quay.io/mojanalytics/grafana:tempFork`](https://quay.io/repository/mojanalytics/grafana?tab=tags) (can't see where the repo for this is)
  - **NOTE**: We'll have to just patch the ingress object manually until we have more clarity on the forked docker image
- [x] Logstash
  - release: `logstash-auth0`
  - ns: `default`
  - chart: `logstash-0.1.1`
  - the chart installed is maintained by us: https://github.com/ministryofjustice/analytics-platform-helm-charts/tree/master/charts/logstash
  - **NOTE**: This helm chart is not installed in `dev`. Is this a problem?
- [X] Concourse
- [X] CP API
- [X] CP Frontend
- [X] AWS Login 🙉 
- [X] Unidler

### DIFF
--- a/charts/cpfrontend/templates/ingress.yaml
+++ b/charts/cpfrontend/templates/ingress.yaml
@@ -9,10 +9,15 @@ metadata:
     kubernetes.io/ingress.class: "nginx"
 spec:
   rules:
-  - host: "cpanel{{- if .Values.Frontend.Branch -}}-{{ .Values.Frontend.Branch }}{{- end -}}.{{ .Values.ServicesDomain }}"
+  - host: {{ template "hostname" . }}
     http:
       paths:
       - backend:
           serviceName: {{ template "fullname" . }}
           servicePort: 80
         path: /
+{{ if .Values.ingress.addTlsBlock }}
+  tls:
+  - hosts:
+    - {{ template "hostname" . }}
+{{ end }}
--- a/charts/aws-login/CHANGELOG.md
+++ b/charts/aws-login/CHANGELOG.md
@@ -1,0 +1,19 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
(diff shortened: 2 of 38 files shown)


charts/cpfrontend/templates/ingress.yaml
```diff
@@ -9,10 +9,15 @@ metadata:
     kubernetes.io/ingress.class: "nginx"
 spec:
   rules:
-  - host: "cpanel{{- if .Values.Frontend.Branch -}}-{{ .Values.Frontend.Branch }}{{- end -}}.{{ .Values.ServicesDomain }}"
+  - host: {{ template "hostname" . }}
```
r4vi: nice refactor

xoen:
FYI: I'm going to do pretty much the same thing in the other charts deployed under `*.services` and push here and rename PR title as it's probably better to keep all these related things together just-in-case.

PR #220, opened 2018-09-28
Remove heapster chart

kerin:
Heapster has been superseded by metrics-server, which has a chart in the main Helm repo. We’re now using that, so the heapster chart is no longer required.


r4vi:
@kerin so do we really want to remove it?

kerin:
@r4vi yep ☠️

PR #222, opened 2018-10-02
Link to Concourse pipeline which updates helm repo

xoen:
So that's it's easy to get there when you're not familiar with how things work.


xoen:
FYI: Merging this little readme update.

PR #224, opened 2018-10-03
Idler 1.0.0 (appVersion: v0.3.0): Support Nanocores

r4vi:
This bumps the image version of the idler to a version that supports nanocores
as a unit/format for cpu usage.

### DIFF
--- a/charts/idler/Chart.yaml
+++ b/charts/idler/Chart.yaml
@@ -1,4 +1,5 @@
 apiVersion: v1
 description: idler cronjob
 name: idler
-version: 0.2.4
+version: 1.0.0
+appVersion: v0.3.0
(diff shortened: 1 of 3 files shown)


charts/idler/Chart.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: idler cronjob
 name: idler
-version: 0.2.4
+version: 0.3.0
```
kerin: `version` should be the version of the _chart_ rather than the application - there's a separate [`appVersion` key](https://github.com/helm/helm/blob/master/docs/charts.md#the-chartyaml-file) for that.

r4vi: @kerin fixed now

PR #225, opened 2018-10-05
JupyterLab & RStudio: Add idleable label

r4vi:
Add `idleable: "true"` so that Jupyter and Rstudio can opt-in to being idled with a single label selector


r4vi:
@kerin because the idler uses prefix `mojanalytics.xyz/` for it's annotations, I've changed these to match

PR #227, opened 2018-10-09
Remove nginx-ingress chart

davidread:
* the charts README points people at using the (official) stable one instead
* I added a version number for install predictability
* the namespace was wrong


r4vi:
:wastebasket:

PR #235, opened 2018-10-16
Control Panel v1.1.1: Support Jupyter [bugfix]

r4vi:
Adds the secrets to the secrets file so that the pods can access the jupyter
lab secrets


xoen:
Looks good to me...but so it did last PR soz😭

PR #236, opened 2018-10-17
Update jupyter images to latest tag

RobinL:
This updates the Jupyter image tag to v0.6.0, which adds geospatial functionality, support for s3 access in spark, and the latest version of jupyter lab.

Might want to pay a little more attention to this PR than usual since it's my first into the helm charts repo!

### DIFF
--- a/charts/jupyter-lab/CHANGELOG.md
+++ b/charts/jupyter-lab/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [0.1.9] - 2018-10-16
+### Changed
+- Updated image tag to v0.6.0, which adds geospatial functionality, support for s3 access in spark, and the latest version of jupyter lab.
+
+
 ## [0.1.8] - 2018-10-05
 ### Added
 - Add `idleable=true` annoation to the deployment. This is to allow opt-in to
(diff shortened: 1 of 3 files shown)


charts/jupyter-lab/CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [0.1.9] - 2018-10-16
```
xoen: 🏆

charts/jupyter-lab/Chart.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: Jupyter Lab with Auth0 authentication proxy
 name: jupyter-lab
-version: 0.1.8
+version: 0.1.9
```
xoen: We don't consider this new version of Jupyter lab breaking compatibility right?

charts/jupyter-lab/values.yaml
```diff
@@ -25,7 +25,7 @@ authProxy:
 
 jupyter:
   image: quay.io/mojanalytics/datascience-notebook
```
xoen: I know it's already there but it's a bit weird that this image is called `datascience-notebook` instead of, let's say, `jupyter-lab`.

charts/jupyter-lab/Chart.yaml
RobinL: Good point.  That's right, we're not breaking compatibility here.

charts/jupyter-lab/values.yaml
RobinL: it's because there are actually two, one called `all-spark` and on called `datascience-notebook` - they are two different flavours of jupyter lab and mirror the naming conventions in the base images we're using

PR #237, opened 2018-10-17
RStudio 1.5.7: Better probes

xoen:
### What
Looks like sometimes auth proxy crashes but container is not terminated (see ticket)

Looking at the RStudio deployment I see that we defined a `readinessProbe` but not a `livenessProbe`, this is likely to be the cause of the problem.

Just a reminder:
- `readinessProbe` determine whether the container should receive traffic
- `livenessProbe` determine if the container is healthy and whether kubernetes will terminate it or not
- `initialDelaySeconds` on `livenessProbe` needs to be enough for the pod to start otherwise k8s will terminate the pod before it can be healthy and you’ll end-up in a loop

#### Changed
- Don't hardcode target port in `Service`, use provided value
- Renamed ports in `Deployment` to be more explicit and avoid any possible
  confusion (they're called `proxyPort` and `rstudioPort` respectively now)
- Increaded probe frequency by reduced `periodSeconds` to `2` from `5`,
  this could help with the user experience as k8s could respond to problems
  more promptly.

#### Added
- Added `livenessProbe` to containers. This will ensure that containers
  are terminated (and restarted) when containers stop serving traffic
  for whatever reason.

### See
- [Pod lifecycle / Container probes](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-probes)
- [Configure Liveness and Readiness Probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes)
- [Kubernetes Health Checks with Readiness and Liveness Probes (Kubernetes Best Practices) [VIDEO]](https://www.youtube.com/watch?v=mxEvAPQRwhw)
### Ticket
https://trello.com/c/92huREQl/1340-fix-liveness-probe-auth-proxy-crash-not-being-handled

### DIFF
--- a/charts/rstudio/CHANGELOG.md
+++ b/charts/rstudio/CHANGELOG.md
@@ -11,6 +11,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Don't hardcode target port in `Service`, use provided value
 - Renamed ports in `Deployment` to be more explicit and avoid any possible
   confusion (they're called `proxyPort` and `rstudioPort` respectively now)
+- Increaded probe frequency by reduced `periodSeconds` to `2` from `5`,
+  this could help with the user experience as k8s could respond to problems
+  more promptly.
+
+### Added
+- Added `livenessProbe` to containers. This will ensure that containers
+  are terminated (and restarted) when containers stop serving traffic
+  for whatever reason.
 
 
 ## [1.5.6] - 2018-10-05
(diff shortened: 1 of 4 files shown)


charts/rstudio/CHANGELOG.md
```diff
@@ -4,24 +4,45 @@ All notable changes to this project will be documented in this file.
...
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+
+## [1.5.7] - 2018-10-17
+
+### Changed
+- Don't hardcode target port in `Service`, use provided value
+- Renamed ports in `Deployment` to be more explicit and avoid any possible
+  confusion (they're now called `proxy` and `rstudio` respectively)
+- Increaded probe frequency by reduced `periodSeconds` to `2` from `5`,
```
davidread: maybe say 'ReadinessProbe' here, now we're talking about 2 probes

xoen: I reduced all probes frequency by reducing how often they're checked but good point as I accidentally forgot an `s`, it should be **probeS**, I'll update 👍

PR #240, opened 2018-10-19
jupyter-lab 0.1.10: Better probes

xoen:
- Added `livenessProbe` to be sure pods are terminated when unhealthy
- Reduced probes' `periodSeconds` from `5` to `2`
- Renamed ports in `Deployment` to `proxy`/`jupyter` instead of using `http`
  for everything which could lead to confusion
- Fixed indentation in `Deployment` YAML file

Ticket: https://trello.com/c/92huREQl/1340-fix-liveness-probe-auth-proxy-crash-not-being-handled

Similar PR: https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/237


xoen:
I'm still seeing some initial failures which I think are caused by the proxy/nodejs being slow at serving requests for some reason I can't really understand (proxy app should be relatively small and straightforward). 

That's why `initialDelay` is at 10s, but it may take a bit longer than from my experiments. I could obviously increase that but that would mean users waiting, etc...it would be best to see why that happens really.

PR #242, opened 2018-10-29
Fix log shipping

djsd123:
Modified image tag from  `0.1.7` to `0.1.8` - See
[Fix-webapp-log-shipping](https://github.com/ministryofjustice/analytics-platform-concourse-github-org-resource/pull/10)


charts/concourse-org-pipeline/values.yaml
```diff
@@ -43,7 +43,7 @@ github:
 
 githubOrgResource:
   image: quay.io/mojanalytics/github-org-resource
-  tag: v0.3.9
+  tag: v0.4.1
```
r4vi: can you update to latest image? 0.5.1? I thought I already did it...

r4vi: I did it in `-config` so not 100% required to bump to 0.5.1 here but it would be nice to keep it in line

PR #243, opened 2018-11-01
Webhook Dispatcher

r4vi:
This is a new chart for deploying a github org hook to concourse pipeline
service. This will allow us to reduce polling frequency for concourse pipelines
because they will now be triggered by a single org webhook


r4vi:
values: https://github.com/ministryofjustice/analytics-platform-config/pull/96

PR #245, opened 2018-11-02
Adds SERVICES_DOMAIN env variable to the pod

davidread:
Needed for https://github.com/ministryofjustice/analytics-platform-control-panel-frontend/pull/294


charts/cpfrontend/README.md
```diff
@@ -35,6 +35,7 @@ The 'kubectl-oidc' Auth0 application should already be created - see: https://gi
 
 | Parameter  | Description     | Default |
 | ---------- | --------------- | ------- |
+| `ServicesDomain` | Domain under which tools are deployed, e.g. `tools.example.com` | `""` |
```
xoen: I know this is just an example in the README but shouldn't this be `services.example.com`?

PR #246, opened 2018-11-07
Websocket Status: 0.1.0

r4vi:
Initial commit of the websocket status page chart. It's quite a simple one as it
needs no configuration.


charts/websocket-status/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+description: Chart for WebsScket Status Page
```
davidread: typo

PR #248, opened 2018-11-13
jupyter-lab 0.1.11: Better probes adjustment

djsd123:
Related to [PR](https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/240)

Intermittently some users are unable to launch Jupyter-Lab successfully
due to probe failures.  There doesn't appear to be a pattern to failing replicas. All we see in the pod
description are the `Liveness/Readiness probes are failing with
getsockopt: connection refused` errors.

The errors are resolved by increasing the values of the probes for
Jupyter or removing the probes entirely.  See [Probe Configuration](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#configure-probes)


djsd123:
No, not at all.  The way you set it before seems to work for 90% of the population.  I suspect a bug in k8s but cannot prove it yet.  Unfortunately, I saw that both probes fail for users affected, otherwise, you're right.

PR #251, opened 2018-11-16
CRAN Proxy Chart

r4vi:
This is a chart for hosting a CRAN proxy in our cluster. The cran proxy can
speed up installation of binary R packages for both users and webapps.

### DIFF
--- a/charts/cran-proxy/values.yaml
+++ b/charts/cran-proxy/values.yaml
@@ -1,0 +1,28 @@
+# Default values for cran-proxy.
+replicaCount: 1
+
+image:
+  name: quay.io/mojanalytics/cran-proxy
+  tag: v0.2.2
+  pullPolicy: IfNotPresent
+
+service:
+  externalPort: 80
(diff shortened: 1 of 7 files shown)


charts/cran-proxy/values.yaml
```diff
@@ -0,0 +1,22 @@
...
+  tag: v0.1.1
+  pullPolicy: IfNotPresent
+
+service:
+  externalPort: 80
+  internalPort: 8000
+  type: ClusterIP
+
+ingress:
+  addTlsBlock: true
```
djsd123: You could reference this with just `{{ .Values.ingress }} as no value or even an empty string is false.

charts/cran-proxy/values.yaml
```diff
@@ -0,0 +1,22 @@
...
+  className: "nginx"
+
+servicesDomain: ""
+
+concourse:
```
r4vi: oops copy & paste... these need to be removed

PR #260, opened 2018-11-27
Remove path from all the ingresses, to match whats deployed to alpha

### LINKED ISSUE #229: Ingresses shouldn't contain paths
Here are a couple of hacks I had to do to make accelerator look like alpha. This is a reminder to myself to apply these somewhere in the set-up.

```
k edit ing cpanel-master-cpanel
# remove the "path: /"
spec:
  rules:
  - host: cpanelapi-master.services.accelerator.data-science.org.uk
    http:
      paths:
      - backend:
          serviceName: cpanel-master-cpanel
          servicePort: 80
        path: /

k edit ing cpfrontend-cpfrontend
# remove the "path: /" and add tls block
  tls:
  - hosts:
    - cpanel-master.services.alpha.mojanalytics.xyz
```

davidread:
Fixes #229

### DIFF
--- a/charts/airflow-k8s/templates/ingress.yml
+++ b/charts/airflow-k8s/templates/ingress.yml
@@ -16,7 +16,6 @@ spec:
       - backend:
           serviceName: {{ .Release.Name }}
           servicePort: 80
-        path: /
   tls:
     - hosts:
       - {{ printf "%s.%s" .Release.Name .Values.toolsDomain | lower }}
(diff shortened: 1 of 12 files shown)


r4vi:
@davidread while this is good for consistency, I don't see why you needed to remove it manually to make it work on accelerator... the default if not specified is `path: /`

davidread:
Yeah, you're probably right.

PR #269, opened 2018-12-14
kube2iam 1.1.2: Added support for regional STS endpoint

xoen:
From : https://github.com/jtblin/kube2iam#aws-sts-endpoint-and-regions

> STS also has regional endpoints which can only be explicitly used
> programatically. The use of a regional sts endpoint can reduce the
> latency for STS requests

By passing the `sts_region` value to the helm chart, kube2iam will use the regional endpoint
which can improve STS requests' latency.

By defaul kube2iam **will not** use regional STS endpoint.


xoen:
FYI: I've [updated the `values.yaml` file](https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/269/files#diff-58907dd3e4cceaf30b67a21936df3a73) to use `jtblin/kube2iam:0.10.4` (official docker image, not ours).

PR #274, opened 2019-01-15
Add Auth Proxy to Kube-Dashboard

djsd123:
[Trello](https://trello.com/c/3pdyjUoO)

The Kubernetes Dashboard is accessible to the public.  This PR puts a
authentication proxy in front of it.


charts/kubernetes-dashboard/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+## [2.0.0] - 2019-01-15
+### Changed
+- Added Auth Proxy to Kubernetes-Dashboard
+- Modified Kubernetes-Dashboard image tag from  `v1.8.3` to `v1.10.1`
+- Switch Kubernetes-Dashboard to insecure backend
```
xoen: (Good stuff on keeping an up-to-date Changelog Mikael 🌟 )

(probably one of my usual stupid question) What's this `--enable-insecure-login=true` flag? Is it just to disable login as we now rely on auth-proxy?

charts/kubernetes-dashboard/templates/deployment.yaml
```diff
@@ -17,32 +17,52 @@ spec:
...
+          - name: AUTH0_DOMAIN
+            valueFrom:
+              secretKeyRef:
+                name: {{ .Chart.Name }}
+                key: auth0_domain
+          - name: AUTH0_CLIENT_ID
+            valueFrom:
+              secretKeyRef:
+                name: {{ .Chart.Name }}
+                key: auth0_client_id
```
xoen: I know we've not been consistent in other charts (probably also my fault), but [Helm's Naming Convenction](https://docs.helm.sh/chart_best_practices/#naming-conventions) is consistent to Go one as far as I can tell so values' keys should be (ideally) be named using camelCase (e.g. `auth0ClientID`, `cookieSecret`, `imagePullPolicy`, etc...)

Obviously not a problem or anything, but probably best to gradually move to that convention if possible.

charts/kubernetes-dashboard/CHANGELOG.md
djsd123: No not stupid. As with many things to do with the dashboard, the flags are cryptic. The flag `--enable-insecure-login=true` puts a login prompt as we had before allowing one to **skip** auth and go in as RO or enter their token to authenticate.  It means we keep the same model as before with the added proxy login.  The insecure referring to `http` instead of `https`.  Have a go: dashboard.services.alpha.mojanalytics.xyz

charts/kubernetes-dashboard/templates/deployment.yaml
djsd123: No worries.  Will change.

djsd123: Done

PR #276, opened 2019-01-16
images-prepuller: Pre-pull/caches images on kubernetes nodes

xoen:
Why?
=====================================================================

Some of our Docker images are quite big and it can take some time for
them to be downloaded.

If that image was not already downloaded on that node, the first user
which spawns a tool using that docker image will have to wait more than
the usual.

This can be confusing for the users as most of the time these
tools starts relatively quickly. When this takes longer the user
perceive this delay as a symptom of "something didn't work".

This is made worse by the fact that we now have configured our
kubernetes cluster to scale up and down. This means that new
nodes could be created more often. These fresh nodes
will not have the most used Docker images yet.

How does it work?
=====================================================================

This helm chart create a DaemonSet (ds) called `prepull`.
It will start a pod with an init container on each of the nodes.
This init container will go through the list of docker images to
cache and will run a `docker pull $IMAGE`.

**NOTE**: As these init containers are, well, containers we need
to mount the node's path where the Docker images are stored
(`/var/run`) in order for the docker images' files to not be
lost when this container terminate.

### DIFF
--- a/charts/images-prepuller/README.md
+++ b/charts/images-prepuller/README.md
@@ -59,3 +59,62 @@ Listing only the required params here. See `/chart-env-config/` for more details
 | ---------- | --------------- | ------- |
 | `docker.images` (required) | **List** of docker images to pre-pull/cache on all nodes. The images need to be in the normal Docker image format, for example `gcr.io/google-containers/busybox:1.27` | `[]` |
 | `docker.hostPath` | This is the path on the node where Docker images are stored. | `"/var/run"` |
+
+
+## How to debug/check images are pulled correctly?
+
+
+### Check if DaemonSet's pods initialised correctly
+If the prepulling failed for whatever reason (most likely a wrong
+image name/tag or a network problem) you'll see that one of the pod
+initialization will fail:
+
+```bash
+$ kubectl -n default get pods  -lname=prepull
+NAME            READY     STATUS       RESTARTS   AGE
+prepull-8nl4f   1/1       Running      0          4m
+prepull-fhj6x   1/1       Running      0          3m
+prepull-m4fdj   0/1       Init:Error   3          48s
+```
+
+### Get nature of failure
+To get more information about the nature of the failure you can
+inspect the logs of the initContainer (called `prepull`):
+
+```bash
+$ kubectl -n default logs -lname=prepull -c prepull
+Pulling images on node 'ip-1-2-3-4.compute.internal'...
+
+Pulling 'failme-really-do-it'...Using default tag: latest
+Error response from daemon: repository failme-really-do-it not found: does not exist or no pull access
+
+[...]
+```
+
+In this specific example the problem was a docker image which
+doesn't exist.
+
+### Determine if images were downloaded
+
+Again, you can inspect the logs to see if an image was downloaded or if it
+was already present on the node:
+
+```bash
+$ kubectl -n default logs -lname=prepull -c prepull
+Pulling images on node 'ip-1-2-3-4.compute.internal'...
+
+Pulling 'gcr.io/google-containers/busybox:1.27'...1.27: Pulling from google-containers/busybox
+Digest: sha256:545e6a6310a27636260920bc07b994a299b6708a1b26910cfefd335fdfb60d2b
+Status: Image is up to date for gcr.io/google-containers/busybox:1.27
+
+
+Pulling 'busybox:1-musl'...1-musl: Pulling from library/busybox
+Digest: sha256:366488474d5b8dfa2546ec5d220e86029925d6c2e54c3fdf45efbfdd06da8e4d
+Status: Downloaded newer image for busybox:1-musl
+
+[...]
+```
+
+In this case you can see that on this node (`ip-1-2-3-4.compute.internal`):
+- `gcr.io/google-containers/busybox:1.27` was already up-to-date
+- `busybox:1-musl` was new and it was downloaded
--- a/charts/images-prepuller/README.md
+++ b/charts/images-prepuller/README.md
@@ -118,3 +118,13 @@ Status: Downloaded newer image for busybox:1-musl
 In this case you can see that on this node (`ip-1-2-3-4.compute.internal`):
 - `gcr.io/google-containers/busybox:1.27` was already up-to-date
 - `busybox:1-musl` was new and it was downloaded
+
+
+## How to determine the images to prepull/cache?
+
+You can have a good idea of from where to start by listing all docker images
+used by all deployments in any namespace:
+
+```bash
+$ kubectl get deployment --all-namespaces -o custom-columns=images:.spec.template.spec.containers[*].image
+```
--- a/charts/images-prepuller/README.md
+++ b/charts/images-prepuller/README.md
@@ -46,9 +46,10 @@ lost when this container terminate.
 To install the chart:
 
 ```bash
-$ helm install charts/images-prepuller -f chart-env-config/ENV/images-prepuller.yml --name images-prepuller --namespace default
+$ helm upgrade --dry-run --debug --install images-prepuller --namespace default charts/images-prepuller -f chart-env-config/ENV/images-prepuller.yml
 ```
 
+**NOTE**: Remove `--dry-run` from command above.
 
 ## Configuration
 
--- a/charts/images-prepuller/README.md
+++ b/charts/images-prepuller/README.md
@@ -1,0 +1,60 @@
+# images-prepuller Helm Chart
+
+Helm chart to pre-pull/cache Docker images on the kubernetes notes.
+
+
+## Why?
+
+Some Docker images are quite big and it can take some time for
+them to be downloaded.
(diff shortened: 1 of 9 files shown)


xoen:
FYI: I didn't have the chance to test this in `dev` yet. I'll do it tomorrow.

charts/images-prepuller/values.yaml
```diff
@@ -0,0 +1,18 @@
+# Default values for images-prepuller.
+
+docker:
+  hostPath: "/var/run"
+  images:
+    - gcr.io/google-containers/busybox:1.27
```
djsd123: Does the image have docker? `$(docker container run -it gcr.io/google-containers/busybox:1.27 sh -c "which docker")`

charts/images-prepuller/templates/configmap.yaml
```diff
@@ -0,0 +1,23 @@
...
+    app: {{ .Release.Name }}
+data:
+  prepull.sh: |
+    #! /usr/bin/env sh
+
+    echo "Pulling images..."
+
+    for image in `cat images.txt`; do
+        echo "\n\nPulling '$image'..."
+        docker pull $image;
```
djsd123: `docker image pull ...` to future proof

charts/images-prepuller/templates/deamonset.yaml
```diff
@@ -0,0 +1,54 @@
...
+          - name: configmap
+            mountPath: /images-prepuller/images.txt
+            subPath: images.txt
+
+      containers:
+        - name: pause
+          image: gcr.io/google_containers/pause
+
+      volumes:
+        - name: docker
```
djsd123: 👍

charts/images-prepuller/templates/configmap.yaml
xoen: Didn't know about that, thanks, I'll update 👍

charts/images-prepuller/templates/deamonset.yaml
```diff
@@ -0,0 +1,54 @@
...
+        name: prepull
+      annotations:
+        # Force DaemonSet's pods to be re-created when config changes
+        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+
+    spec:
+      initContainers:
+        - name: prepull
+          image: docker
+          command: ["docker", "pull", "hello-world"]
```
xoen: Doh! I added the script to the init container but now by looking at it with a fresh brain I realise that I didn't run the script in the init container 🤣

I'll change.

charts/images-prepuller/README.md
```diff
@@ -0,0 +1,60 @@
...
+## How does it work?
+
+This helm chart create a DaemonSet (ds) called `prepull`.
+It will start a pod with an [init container] on each of the nodes.
+This init container will go through the list of docker images to
+cache and will run a `docker pull $IMAGE`.
+
+This is based on this [gist].
+
+**NOTE**: As these init containers are, well, containers we need
```
r4vi: could it be a normal container with `restart: Never`?

charts/images-prepuller/templates/deamonset.yaml
```diff
@@ -0,0 +1,54 @@
...
+          volumeMounts:
+          - name: docker
+            mountPath: /var/run
```
r4vi: maybe just mount `/var/run/docker.sock` to restrict surface area

charts/images-prepuller/templates/deamonset.yaml
```diff
@@ -0,0 +1,54 @@
...
+      labels:
+        name: prepull
+      annotations:
+        # Force DaemonSet's pods to be re-created when config changes
+        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+
+    spec:
+      initContainers:
+        - name: prepull
+          image: docker
```
xoen: I tempted to to pin the version of docker here as well to avoid weird behaviour in the future.

I'll update.

charts/images-prepuller/README.md
xoen: I was making some tweaks and I was **literally** about to mention the lift chat 🏋️ we had.

Yeah I think it may be a good idea. I'll play with it.

Positive:
- we have 1 less (small) container running
- we don't have to pull the `gcr.io/google_containers/pause` image (which is very small, I think less than 100KB)

Negative:
- would be confusing that the DaemonSet doesn't have any running container? It is usually a sign of "something went wrong"

I want to try anyway and I'll probably change.

charts/images-prepuller/templates/deamonset.yaml
xoen: Can you just mount a single file/socket?

I'll try that as well, as I mentioned I'd prefer the container to only have access to the minimum amount of stuff that it strictly needs.

charts/images-prepuller/templates/deamonset.yaml
xoen: https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/276/commits/a4616a95ead01576719a99ae9a36f81cfa91dc8e

charts/images-prepuller/templates/deamonset.yaml
r4vi: look at `subPath` next to `mountPath`

charts/images-prepuller/templates/deamonset.yaml
xoen: https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/276/commits/d6729cd372d34db3c3746ee2a8822bcf738f042e

charts/images-prepuller/templates/configmap.yaml
xoen: https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/276/commits/1a1dc0e895ea4684e75719b846caaaa1f9e2a016

charts/images-prepuller/templates/deamonset.yaml
```diff
@@ -33,7 +33,8 @@ spec:
           command: ["/images-prepuller/prepull.sh"]
           volumeMounts:
           - name: docker
-            mountPath: /var/run
+            mountPath: /var/run/docker.sock
```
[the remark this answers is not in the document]
xoen: I had a look at the official documentation linked above but from the example it wasn't clear if you could just mount a single file.

But this SO question seems to suggest you can do it this way:

```yaml
volumeMounts:
- name: docker
  mountPath: /var/run/docker.sock
  subPath: docker.sock
```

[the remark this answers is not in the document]
xoen: FYI: I'm getting a couple of small errors but I can't fix them today, so they'll have to wait tomorrow:

1. Specifying the docker image tag cause an error for some obscure reason
2. Can't exec the `/images-prepuller/prepull.sh` script. I'll need to give execution permission to that file. Only had a quick look at k8s reference but couldn't find a way in the `volumeMount` specification. **There must be!**, so I'll have another look tomorrow (worst case scenario I'll have to manually do a `$ chmod a+x /images-prepuller/prepull.sh` which is a bit ugly)

charts/images-prepuller/README.md
xoen: FYI: I had a look at [the documentation](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/#pod-template) and it doesn't look is possible.

> A Pod Template in a DaemonSet must have a RestartPolicy equal to Always, or be unspecified, which defaults to Always.

So I'll keep the "pause" container.

djsd123: Fair enough.  I wonder whether you get a pause container running behind a pause container?  Might have to jump on to the node to find out 🤔

charts/images-prepuller/templates/deamonset.yaml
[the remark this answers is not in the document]
xoen: Fix for permission problem: ee0f78928d4af3cd9619afe6ee3fbd7b81503992

charts/images-prepuller/templates/deamonset.yaml
xoen: Also, managed to pin `docker` image's tag to `18.09.1`: 9f699932f98677d8e6d9c6fe305ff22babda14b5

charts/images-prepuller/README.md
xoen: @djsd123 that sounds like the plot of Inception to me 🤣

xoen:
FYI: I'm happy with this now but have another look.

Also, you can easily see the logs of this if you're curious by running the following (on `dev` cluster):

```bash
$ kubectl -n default logs -lname=prepull -c prepull
```

charts/images-prepuller/README.md
djsd123: 😆

PR #277, opened 2019-01-17
Add host label to webapp deployment, ingress and service

andyhd:


r4vi:
@andyhd you'll need to retroactively have to apply this to existing apps or are you waiting for them to just get redeployed?

andyhd:
Yeah, one or the other - i might just redeploy all webapps myself manually when I want to make use of the label

andyhd:
Ah, good idea - I use `kubectl label` and `kubectl annotate` in a shell script I use for testing the unidler.

PR #278, opened 2019-01-24
rbac: Added "generic-support" group with read-only access to main resources

xoen:
These users will be able to read/list the main kubernetes resources:
  - ingresses
  - services
  - deployments
  - pods
  - cronjobs
  - jobs

These users will have these permissions across the whole cluster, so
in all namespaces.

**NOTE**: Users will not be able to edit these resources (or at least
not at this stage) and they will not be able to read secrets or
configmaps.

Ticket: https://trello.com/c/LsQcIL12


xoen:
Cheers, merging and re-deploying to dev/alpha 👍

PR #280, opened 2019-01-25
rbac: Improved install command in README and fixed value key typo

xoen:
- added `--namespace` param to install/upgrade command so that
  if the user is on a different namespace they'll not accidentally
  install the helm chart in the wrong namespace
- added `--dry-run` flag, so that users will need to check helm
  output before actually installing the helm chart
- fixed typo in `appSupport` value: `desciprtion` => `description`

### DIFF
--- a/charts/rbac/README.md
+++ b/charts/rbac/README.md
@@ -11,8 +11,11 @@ kubectl create ns apps-prod
 ```
 Now install the chart:
 ```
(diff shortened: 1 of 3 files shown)


xoen:
FYI: @djsd123 I just noticed a typo in one of the values so fixed that and improved the README a little. Merging this (no need to review)

djsd123:
@xoen 👍

PR #281, opened 2019-01-25
Helm charts security/tidying love ❤️

xoen:
I did some auditing on our helm charts.

Some helm charts put sensitive data directly into deployment/pod/cronjob/job definition. Sensitive stuff should go in secrets (so that if we grant someone permission to read ConfigMaps/pods we're confident there is anything sensitive there)

- `airflow` moved OAuth credentials from Configmap to Secret
- `airflow` moved admin credentials into Secret from init container definition- Removed old/unused `shiny-app` helm chart (we now use `webapp` for all sort of web applications deploys)


charts/airflow-k8s/templates/secrets.yml
```diff
@@ -13,3 +13,23 @@ data:
   secret-key: {{ printf .Values.airflow.secretKey | b64enc | quote }}
   smtp-user: {{ printf .Values.airflow.config.smtp.smtp_user | b64enc | quote }}
   smtp-password: {{ printf .Values.airflow.config.smtp.smtp_password | b64enc | quote }}
+
+  admin-username: {{ printf .Values.airflow.admin.username | b64enc | quote }}
+  admin-password: {{ printf .Values.airflow.admin.password | b64enc | quote }}
+  admin-email: {{ printf .Values.airflow.admin.email | b64enc | quote }}
+
+  client_secret.json: |
```
xoen: FYI: I'm an idiot. This will need to be base64 encoded of course, so I'll have to find a way to transform this blob into a nice base64 string 😓.

(Currently I'm getting a `v1.Secret.Data: decode base64: illegal base64 data at input byte 0, error found in #10 byte of [...]`)

xoen: Found an elegant workaround thanks to @alkar - there is this beautiful property that `Secretz` have called [`stringData`](https://kubernetes.io/docs/concepts/configuration/secret/#creating-a-secret-manually) which is syntactic sugar for defining secrets without having to Base64-encode them.

See commit: https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/281/commits/94e0e447c318193ef39c468bb6ad8e2e9c4b17cd

xoen:
FYI: Tested in `dev` and it seems to work. I'll test it in `alpha`/break things tomorrow morning.

charts/airflow-k8s/templates/secrets.yml
djsd123: 😍

PR #282, opened 2019-01-29
Idler/unidler with status and reduced race-condition probability

andyhd:


charts/idler/values.yaml
```diff
@@ -12,7 +12,7 @@ schedule: "0 22 * * *"
 # See https://kubernetes.io/docs/user-guide/cron-jobs/#concurrency_policy
 concurrencyPolicy: Forbid
 
-labelSelector: 'app=rstudio'
+labelSelector: 'app'
```
xoen: Does it mean that idler will idle pretty much anything with an `app` label? Is this what we want? That is **VERY BROAD** and could cause of sort of unexpected problems.

Bear in mind I may miss something here.

andyhd: My mistake - don't know why I did that.

xoen: No worries.

PR #290, opened 2019-02-05
Prometheus Chart Migration

djsd123:
Relates to: [PR](https://github.com/ministryofjustice/analytics-platform-config/pull/114)

- Initial commit
- Migrated our deployment from [CoreOS's deprecated chart](https://github.com/coreos/prometheus-operator/tree/master/helm) to [Helm stable repo chart](https://github.com/helm/charts/tree/master/stable/prometheus-operator)
- Secured [Prometheus](https://prometheus.services.dev.mojanalytics.xyz) and [AlertManager](https://alertmanager.services.dev.mojanalytics.xyz) front-ends with [auth-proxy](https://github.com/ministryofjustice/analytics-platform-auth-proxy)
- Migrated Dashboard templates to Grafana 5.4.x spec
- Migrated Rules, Alerts and Dashboards from the previous deployment
- Added `PodSecurityPolicy` in preparation for [Trello](https://trello.com/c/0mfKXhjp)

### DIFF
--- a/charts/prometheus-resources/README.md
+++ b/charts/prometheus-resources/README.md
@@ -1,0 +1,91 @@
+# Prometheus-Resources
+
+Deploys resources related to the [Prometheus-Operator](https://github.com/helm/charts/tree/master/stable/prometheus-operator) helm chart
+
+This chart:
+- Adds our custom prometheus rules 
+- Secures the [Prometheus](https://prometheus.services.dev.mojanalytics.xyz) and [AlertManager](https://alertmanager.services.dev.mojanalytics.xyz) front-ends with [auth-proxy](https://github.com/ministryofjustice/analytics-platform-auth-proxy) 
+
+## Dependencies/Prerequisites 
+
+- This Chart requires that the [Prometheus-Operator](https://github.com/helm/charts/tree/master/stable/prometheus-operator) helm chart be deployed first
+- Helm version 2.10+
+- Kubernetes 1.10+ with Beta APIs
+
+## Installing and Deleting
+
+**Note**:
+
+There are some issues with race conditions pertaining to Custom Resource Definition hooks used by the [Prometheus-Operator](https://github.com/helm/charts/tree/master/stable/prometheus-operator) helm chart. 
+This prevents the [Prometheus-Operator](https://github.com/helm/charts/tree/master/stable/prometheus-operator) helm chart from being installed directly (__with the helm cmd__) 
+and indirectly (__with__ `requirements.yaml`).
+
+There is a [fix](https://github.com/helm/helm/pull/5112). However, this is not expected until __Helm version 2.13.x__.
+
+The steps to install a complete deployment are listed below.  
+
+
+**Before Proceeding**
+Ensure that the following value is set to `false`. (Relating to [bug](https://github.com/helm/helm/issues/4925) mentioned above)
+
+```yaml
+prometheusOperator:
+  createCustomResource: false
+```
+
+### Install Commands
+
+Install [Custom Resource Definitions](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/):
+```bash
+kubectl apply -f https://raw.githubusercontent.com/coreos/prometheus-operator/master/example/prometheus-operator-crd/alertmanager.crd.yaml
+kubectl apply -f https://raw.githubusercontent.com/coreos/prometheus-operator/master/example/prometheus-operator-crd/prometheus.crd.yaml
+kubectl apply -f https://raw.githubusercontent.com/coreos/prometheus-operator/master/example/prometheus-operator-crd/prometheusrule.crd.yaml
+kubectl apply -f https://raw.githubusercontent.com/coreos/prometheus-operator/master/example/prometheus-operator-crd/servicemonitor.crd.yaml
+```
+
+Install [Prometheus-Operator](https://github.com/helm/charts/tree/master/stable/prometheus-operator) helm chart:
+```bash
+helm install --name prometheus-operator stable/prometheus-operator -f chart-env-config/<ENV>/stable-prometheus.yaml --namespace=monitoring --no-crd-hook
+```
+**Note** the `--no-crd-hook`
+
+
+Install this chart:
+```bash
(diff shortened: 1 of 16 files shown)


charts/prometheus-resources/install-chart.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
+
+set -ue
+
+CRDS=("https://raw.githubusercontent.com/coreos/prometheus-operator/master/example/prometheus-operator-crd/alertmanager.crd.yaml"
```
xoen: Are these "examples" or these resources considered stable and we can rely on them?

charts/prometheus-resources/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+**Note**:
+
+There are some issues with race conditions pertaining to Custom Resource Definition hooks used by the [Prometheus-Operator](https://github.com/helm/charts/tree/master/stable/prometheus-operator) helm chart. 
+This prevents the [Prometheus-Operator](https://github.com/helm/charts/tree/master/stable/prometheus-operator) helm chart from being installed directly (__with the helm cmd__) 
+and indirectly (__with__ `requirements.yaml`).
+
+There is a [fix](https://github.com/helm/helm/pull/5112). However, this is not expected until __Helm version 2.13.x__.
+
+The steps to install a complete deployment are listed below.  However, there is a simple [script](./install-chart.sh) included to save you the trouble until this [bug](https://github.com/helm/helm/issues/4925) has been confirmed fixed.
```
xoen: Is that script necessary if steps to install CRDs and chart are documented below? It may be confusing as someone may not be sure which path to take, but maybe it's just me :D

charts/prometheus-resources/README.md
```diff
@@ -0,0 +1,91 @@
...
+```bash
+kubectl apply -f https://raw.githubusercontent.com/coreos/prometheus-operator/master/example/prometheus-operator-crd/alertmanager.crd.yaml
+kubectl apply -f https://raw.githubusercontent.com/coreos/prometheus-operator/master/example/prometheus-operator-crd/prometheus.crd.yaml
+kubectl apply -f https://raw.githubusercontent.com/coreos/prometheus-operator/master/example/prometheus-operator-crd/prometheusrule.crd.yaml
+kubectl apply -f https://raw.githubusercontent.com/coreos/prometheus-operator/master/example/prometheus-operator-crd/servicemonitor.crd.yaml
+```
+
+Install [Prometheus-Operator](https://github.com/helm/charts/tree/master/stable/prometheus-operator) helm chart:
+```bash
+helm install --name prometheus-operator stable/prometheus-operator -f chart-env-config/<ENV>/stable-prometheus.yaml --namespace=monitoring --no-crd-hook
```
xoen: The more I work with helm chart and update existing helm charts the more I'm convinced that in README, etc we should try to stick with the `helm upgrade --dry-run --debug --install ...` version:
1. `upgrade --install` is idempotent so we can re-run it even it we already installed the helm chart (which often is the case when developing or even testing things)
2. the `dry-run` is a bit sneaky unless we document it but IMHO is always a good practice to preview what you're about to install (and often is educational) before you actually go ahead.

charts/prometheus-resources/README.md
djsd123: Sure, I can scrap the script if we think it muddies the waters.  I agree with you though 🥇

xoen: Up to you. I had a similar situation when I was trying to make Istio work and to be honest at first I was confused by their direction and the different paths depending on the version of Helm chart used, etc...

Personally - until helm/k8s handle CRDs better - I prefer documenting that you need to install the CDRs beforehand and that should be clear enough (to me at least). But it really depends on how much experience the person installing this stuff has with CRDs, etc...probably in the case of this helm chart I'd be the only one getting confused :)

charts/prometheus-resources/README.md
djsd123: Agreed, `upgrade` is better. 👍  We do `install` deliberately here to work around bugs with the `CRD's and the associated hooks.

Agreed `--dry-run` is good/ok. 👍 I only ever use that flag when I am doing a large deployment. i.e. Updating all our webapps and other user resources in place.  Personally, I don't often use it because it never raises any issues that often occur on the `tiller` end like `dry-runs` with other with other tools.  I'll add it if we agree it's best.

charts/prometheus-resources/README.md
djsd123: Done

charts/prometheus-resources/README.md
xoen: Don't have to. I was mainly thinking aloud.

As said, `--dry-run` is sneaky so unless we have a big flashing **NOTE** users could not realise it was a dry-run and didn't actually do anything if distracted. Up to you really.

charts/prometheus-resources/templates/ingress.yaml
```diff
@@ -0,0 +1,26 @@
...
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: {{ $.Release.Name }}-{{ .name }}
+  namespace: {{ $.Release.Namespace }}
+  labels:
+    app: {{ .name }}-{{ $.Values.authProxy.name }}
+    target: {{ .name }}
+  annotations:
+    kubernetes.io/ingress.class: nginx
```
r4vi: maybe make this optional / configurable. What if we want to use another ingress at some point

charts/prometheus-resources/values.yaml
```diff
@@ -0,0 +1,171 @@
...
+  containerPort: 3000
+  resources:
+    limits:
+      cpu: 100m
+      memory: 128Mi
+    requests:
+      cpu: 25m
+      memory: 64Mi
+  cookie_maxage: 36000 #10 hours
+  cookie_secret: "tony-montana"
```
r4vi: :laughing:

charts/prometheus-resources/templates/ingress.yaml
djsd123: Good idea.  Done!

djsd123:
This is a big chart with lots going on and is hard to review, so cheers for your efforts @r4vi and @xoen.

PR #293, opened 2019-02-15
Rstudio v1.6.0: Switch to conda rstudio image

r4vi:
- Version of RStudio image with conda (r3.5.1-py3.7-conda)
- it doesn't point at the cran proxy
- does initial setup in an init container
- has a rstudio conda environment
- maintains support for projects still using packrat
- can point to an older image that doesn't support init by setting `rstudio.image.init` to false

### DIFF
--- a/charts/rstudio/CHANGELOG.md
+++ b/charts/rstudio/CHANGELOG.md
@@ -4,6 +4,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [1.6.0] - 2019-02-14
+
+### Changed
+- Version of RStudio image with conda (r3.5.1-py3.7-conda)
+- it doesn't point at the cran proxy
+- does initial setup in an init container
+- has a rstudio conda environment
+- maintains support for projects still using packrat
+
 ## [1.5.8] - 2018-10-17
 
 ### Changed
--- a/charts/rstudio/Chart.yaml
+++ b/charts/rstudio/Chart.yaml
@@ -1,4 +1,5 @@
 apiVersion: v1
 description: RStudio with Auth0 authentication proxy
 name: rstudio
-version: 1.5.8
+version: 1.6.0
+appVersion: "RStudio: 1.1.463+conda, R: 3.5.1, Python: 3.7"
--- a/charts/rstudio/templates/deployment.yml
+++ b/charts/rstudio/templates/deployment.yml
@@ -21,6 +21,26 @@ spec:
         iam.amazonaws.com/role: {{ .Values.aws.iamRole }}
     spec:
       serviceAccountName: {{ .Values.username }}-rstudio
+      {{ if and .Values.rstudio.image.init }}
+      initContainers:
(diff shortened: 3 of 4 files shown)


r4vi:
I have 2-3 users on this helm chart and it seems good

charts/rstudio/CHANGELOG.md
```diff
@@ -4,6 +4,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [1.6.0] - 2019-02-14
+
+### Changed
+- Version of RStudio image with conda (r3.5.1-py3.7-conda)
```
xoen: Nice tag 👍

charts/rstudio/Chart.yaml
```diff
@@ -1,4 +1,5 @@
 apiVersion: v1
 description: RStudio with Auth0 authentication proxy
 name: rstudio
-version: 1.5.8
+version: 1.6.0
+appVersion: 3.5.1
```
xoen: This is the version of R right? Do we care about the version of RStudio. I mean, this is one of the most confusing bit of packaging an IDE and 2 languages I suppose.

The image tag has R version and Python version which is great...maybe we don't care about RStudio version itself, we probably care less than the version of the languages.

I like that we have `appVersion` as it's useful to debug gremlins.
I wonder if in this case we can bend this field to the limits and have a quoted string like:

`appVersion: "RStudio: 20190301+conda, R: 3.5.1, Python: 3.7"`

Assuming Helm devs didn't take any "design decision" here and that would work, it may still be overkill...we could use the docker image tag as "app version", especially now that include both R and Python version. But maybe that would be redundant?

Just thinking aloud, sorry.

charts/rstudio/templates/deployment.yml
```diff
@@ -21,6 +21,26 @@ spec:
...
+      {{ if and .Values.rstudio.image.init }}
+      initContainers:
+        - name: r-studio-server-init
+          image: "{{ .Values.rstudio.image.repository }}:{{ .Values.rstudio.image.tag }}"
+          imagePullPolicy: {{ .Values.rstudio.image.pullPolicy }}
+          command:
+            - /usr/local/bin/start.sh
+            - init
+          env:
+            - name: USER
```
xoen: Just had a thought. I wonder if we should use the [`user-secrets` in the user namespace](https://github.com/ministryofjustice/analytics-platform-helm-charts/blob/master/charts/init-user/templates/user-secrets.yml) more heavily in the future...

e.g. we may not even need to pass the `username` as value here, it would be in the secret. Similarly, we may put the `AWS_DEFAULT_REGION` in these secrets and just use it when we install the chart/the deployment get these values from the user secrets.

Anyway, sorry, another random thought for the future maybe.

charts/rstudio/Chart.yaml
r4vi: I like this

charts/rstudio/templates/deployment.yml
r4vi: leaving this for the future

xoen: Of course, it goes without saying :)

PR #296, opened 2019-02-19
airflow-k8s 0.1.1: Fix for some requests made over **non-secure** HTTP

xoen:
- Fixed problem where some of the authentication requests were done over
  **non-secure** HTTP connection.
- added `airflow.config.webserver.enable_proxy_fix` which
  by default is `True` but can be changed if necessary
- See:
  - Airflow issue: [`AIRFLOW-3137`](https://issues.apache.org/jira/browse/AIRFLOW-3137)
  - Airflow fix: [`PR #3983`](https://github.com/apache/airflow/pull/3983)

**NOTE**: You should only enable the `ProxyFix` middleware when running
Airflow behind a trusted proxy (AWS ELB, nginx, etc.).

Ticket: https://trello.com/c/6AbABA6a/161-make-sure-airflow-is-always-using-https

### DIFF
--- a/charts/airflow-k8s/CHANGELOG.md
+++ b/charts/airflow-k8s/CHANGELOG.md
@@ -5,6 +5,24 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+## [0.1.1] - 2019-02-19
+### Security
+- Fixed problem where some of the authentication requests were done over
+  **non-secure** HTTP connection.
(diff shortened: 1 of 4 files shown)


xoen:
FYI: I'm about to test this in `dev` but I can't see why this shouldn't work (I manually tested the fix earlier in `dev` and it works)

xoen:
FYI: I've updated `dev` Auth0 client to remove the HTTP URLs (non-HTTP) and some "crap" (e.g. airflow2 URL).

xoen:
FYI: Tested in `dev` (as said Airflow Auth0 client has now only HTTPs URLs in `dev`), and it works as expected 🎊 

Merging and releasing in `alpha`, thanks for the review!

PR #297, opened 2019-02-25
airflow-k8s 0.2.0: Upgraded Redis and using NetworkPolicy

xoen:
- Upgraded `redis` chart dependency `3.6.4` => `6.1.3`
  - This upgrade `redis` version `4.0.10` => `4.0.13`
- Updated values to reflect helm chart upgrade
- Enabled `NetworkPolicy` so that only Airflow webserver
  can access redis
- Only download Redis docker image if it's not already present
  in the k8s node (`IfNotPresent`).
  Image tag is not `latest` so no reason to make unnecessary
  requests and waste time.

Ticket: https://trello.com/c/m6m3onzw/160-airflow-helm-chart-upgrade-redis-dependency


xoen:
💺💺chairs mait 🤗

PR #299, opened 2019-03-01
Add concourse fly creds to webhook dispatcher so it can use fly

s-block:


charts/webhook-dispatcher/values.yaml
```diff
@@ -20,3 +20,7 @@ servicesDomain: ""
 concourse:
   base_url: ""
   token: ""
+  admin_username: ""
```
r4vi: does it need admin? I'm guessing it won't be triggering pipelines in the admin team

s-block: Not sure. team is passed from source so I wasn't 100% sure. Can remove it if you think it isn't or won't be needed in the future. Would the new lookup table pipelines go in to main as well?

r4vi: everything goes in main unless it's something for us that a user won't care about seeing

r4vi: I don't mind either way... just asking; already approved

s-block: Removed admin creds

PR #95, opened 2018-02-14
cpfrontent 1.3.4: Removed `APP_PORT` environment variable

xoen:
See related PR: https://github.com/ministryofjustice/analytics-platform-control-panel-frontend/pull/131

PR #305, opened 2019-03-13
unidler v3.2.0: Fix for `undefined` error

xoen:
New version of unidler which should fix [Issue #95]:

"Unidler giving misleading messages", specifically it sometimes shows a
strange `undefined` JS error. We think this was caused by the unidler
erroneously updating the app Service too early. If then the client
for whatever reason makes another request to `/events` this will go to
the actual app which doesn't have a `/events` endpoint.

Other improvements:
- users receive more updates as their app is being unidled
- fixed space between text and image
- unidler logs more useful things now
- other internal refactorings which shouldn't be user-facing
- Don't set `INGRESS_CLASS_NAME` variable as unidler doesn't use it anymore

[Issue #95]: https://github.com/ministryofjustice/analytics-platform/issues/95


xoen:
Thanks @r4vi 🌟

PR #308, opened 2019-03-14
FluentD Resources WebApp

djsd123:
It appears that the increased cpu shares for webapps has not been enough
to settle the throttling alerts.  I’m going to increase the limits
further to 1 core.

- Increased cpu shares for fluentd. Requests and limits from 200m to 500m and 500m to 1 core respectively


xoen:
👨‍🚒

PR #311, opened 2019-03-18
FluentD Pod Priority Class

djsd123:
[Trello](https://trello.com/c/Ay2ryNvC)

- Added `priorityClassName` reference to system `PriorityClass`
- Increased burstable cpu limit from `500m` to 1 cpu
- Changed apiVersion for `daemonset`

### DIFF
--- a/charts/fluentd/values.yaml
+++ b/charts/fluentd/values.yaml
@@ -11,11 +11,12 @@ elasticsearch:
   logstash_prefix: "logstash"
 resources:
   limits:
-    cpu: 500m
+    cpu: 1000m
     memory: 1Gi
   requests:
     cpu: 200m
     memory: 512Mi
+priorityclassname: system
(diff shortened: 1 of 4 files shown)


charts/fluentd/values.yaml
```diff
@@ -11,11 +11,12 @@ elasticsearch:
   logstash_prefix: "logstash"
 resources:
   limits:
-    cpu: 500m
+    cpu: 1000m
     memory: 1Gi
   requests:
     cpu: 200m
     memory: 512Mi
+priorityclassname: system
```
xoen: 😎

xoen: Annoying late comment as usual...shouldn't the case be `priorityClassName`? Don't worry of course, maybe for next time you touch this chart and the config :)

PR #316, opened 2019-03-21
unidler v3.2.2: Less errors reaching the user and UX tweaks

xoen:
- Removed #/6 "progress" messages as confusing (for some users)
- Better handling of "nonexistent key" error when patching Service to avoid
  showing "Failed to redirect back your app" when we can safely recover.
- when unidling an app which has annotation with unidled number of replicas
  of `0`, assume this is a mistake and set number of replicas to `1`.

See PRs:
- https://github.com/ministryofjustice/analytics-platform-go-unidler/pull/7
- https://github.com/ministryofjustice/analytics-platform-go-unidler/pull/8


xoen:
Thanks @djsd123 🌟 

FYI: Tested in `dev` and merging/deploying to `alpha` soon.

PR #318, opened 2019-03-27
unidler v4.0.0: Bumped unidler version to `v1.0.0`

xoen:
**NOTE**: This is a breaking change as the unidler now uses the `host` labels
to retrieve the resources to change. But these labels need to be there.
So not exactly retro-compatible.

- use `host` labels to find Ingress, Deployment, etc...
- use `mojanalytics.xyz/replicas-when-unidled` annotation instead of old
  `mojanalytics.xyz/idled-at` which was trying to do too many things at once

**SEE**:
- [unidler changes](https://github.com/ministryofjustice/analytics-platform-go-unidler/pull/9)
- [migration to add labels](https://github.com/ministryofjustice/analytics-platform-ops/pull/266)

**Ticket**: https://trello.com/c/NMpN8B1j/132-add-host-label-to-all-idleable-apps-tools

### DIFF
--- a/charts/unidler/CHANGELOG.md
+++ b/charts/unidler/CHANGELOG.md
@@ -5,6 +5,23 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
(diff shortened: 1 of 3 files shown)


xoen:
FYI: Tested in `dev` and `alpha` and it works as expected. (released this in both `dev` and `alpha`).

PR #320, opened 2019-03-28
WebApp: FluentD Upgrade

djsd123:
Upgrading image to make use of better error handling features, log
output and compatibility with elasticsearch `6.x`.

- Setting target `target_type_key` is more explicit and means the
elasticsearch plugin won't need to fall through the stack or error out to determine the
indexing type
- __Newwer Feature__: `log_es_400_reason` means we get log output on why
Elasticsearch returned a `400: bad request` without having to set log to
`DEBUG`
- Removed record transformer `types` key because it's not needed

### DIFF
--- a/charts/webapp/Chart.yaml
+++ b/charts/webapp/Chart.yaml
@@ -1,4 +1,5 @@
 apiVersion: v1
 description: webapp Helm chart for Kubernetes
 name: webapp
-version: 1.3.22
+fluentdVersion: 2.5.0
+version: 1.3.23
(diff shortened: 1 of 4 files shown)


charts/webapp/Chart.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: webapp Helm chart for Kubernetes
 name: webapp
-version: 1.3.22
+version: 1.3.23
```
xoen: It may be a good idea to also have `appVersion` here.

djsd123: Fluentd is the sidecar in this scenario so I would say no.  However, I added `fuentdVersion`

xoen: You're right. In the case of the helm chart `appVersion` doesn't make much sense, sorry 😬

PR #321, opened 2019-03-28
Jupyter-lab: Upgrade auth proxy image

r4vi:
upgrade to auth proxy that supports _tunneling_. Add environment variable to the
deployment to enable the feature.

### DIFF
--- a/charts/jupyter-lab/templates/deployment.yaml
+++ b/charts/jupyter-lab/templates/deployment.yaml
@@ -66,6 +66,10 @@ spec:
               value: {{ .Values.Username }}
             - name: COOKIE_MAXAGE
               value: {{ .Values.authProxy.cookie_maxage | quote }}
+            - name: TUNNEL_ENABLED
+              value: {{ .Values.authProxy.tunnel.enabled | quote }}
+            - name: TUNNEL_PORT_RANGE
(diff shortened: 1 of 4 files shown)


r4vi:
change to platform guidance incoming to explain how to use this (just going to record some animated gifs)

charts/jupyter-lab/templates/deployment.yaml
```diff
@@ -66,6 +66,10 @@ spec:
               value: {{ .Values.Username }}
             - name: COOKIE_MAXAGE
               value: {{ .Values.authProxy.cookie_maxage | quote }}
+            - name: TUNNEL_ENABLED
+              value: {{ .Values.authProxy.tunnel.enabled | quote }}
+            - name: TUNNEL_PORT_RANGE
+              value: {{ .Values.authProxy.tunnel.port_range | quote }}
```
xoen: Sorry to be late to the party. Helm convention is to use camelCase for Values, e.g. `.Values.authProxy.tunnel.portRange`. Don't know if you're still in time to change.

r4vi: yeah but I'm keeping consistent with the other values which are bad_cased

xoen: 😆

PR #322, opened 2019-03-29
idler 2.0.0: bumped idler tag to `v1.0.0`

xoen:
- Use `idler` to `v1.0.0`:
  - adds `mojanalytics.xyz/replicas-when-unidled` annotation with number of
    replicas desired at the time of udling. This is used by `unidler`
  - `mojanalytics.xyz/idled-at` only contains time when app was idled,
    nothing else
  - renamed env var `RSTUDIO_ACTIVITY_CPU_THRESHOLD` =>
    `CPU_ACTIVITY_THRESHOLD`
  - internal refactorings
  - `labelSelector` defaults to `mojanalytics.xyz/idleable=true` now
    (in `idler`)
  - See PR: ministryofjustice/analytics-platform-idler#68
- added `app=$RELEASE_NAME` to job's pods so that we can see logs and
  investigate things much more easily
- `labelSelector` defaults to `mojanalytics.xyz/idleable=true` now (in helm
  chart)
- `logLevel` defaults to `DEBUG`. We want to see what's going on unless
  we explicit silence it for whatever reason.

### DIFF
--- a/charts/idler/CHANGELOG.md
+++ b/charts/idler/CHANGELOG.md
@@ -4,22 +4,48 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+
+## [2.0.0] - 2019-03-29
+### Changed
+- Use `idler` to `v1.0.0`:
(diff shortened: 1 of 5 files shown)


xoen:
Thanks @r4vi - not trying to be sneaky but I also just updated the README: https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/322/commits/66b7f114bafe4f16357bb8afdbee0fb08caf9c68 (you can probably ignore that but as always feel free to have a look)

PR #326, opened 2019-04-08
Add airflow-sqlite helm chart

s-block:
Add a helm chart for airflow-sqlite.

This is so the users can test dags running as their own user before merging and releasing them.

This will be run on demand by users via the control panel. The chart will run airflow using sqlite as the database in the users kubernetes namespace.

The users home directory will be mounted and airflow will use `/home/username/airflow/dags` as the dags directory.

### DIFF
--- a/charts/airflow-sqlite/templates/webserver-deployment.yml
+++ b/charts/airflow-sqlite/templates/webserver-deployment.yml
@@ -1,0 +1,106 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ .Release.Name }}-webserver
+  namespace: {{ .Release.Namespace }}
+  labels:
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    app: {{ .Release.Name }}
+    component: webserver
+  annotations:
+    checksum/config: {{ include (print $.Template.BasePath "/configmap.yml") . | sha256sum }}
+spec:
+  selector:
+    matchLabels:
+      app: {{ .Release.Name }}
+      component: webserver
+  replicas: 3
+  template:
+    metadata:
+      labels:
+        app: {{ .Release.Name }}
+        component: webserver
+        airflow-redis-client: "false"
+    spec:
+      serviceAccountName: {{ .Release.Name }}
+      containers:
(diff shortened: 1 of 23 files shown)


charts/airflow-sqlite/templates/webserver-deployment.yml
```diff
@@ -0,0 +1,106 @@
...
+                secretKeyRef:
+                  name: {{ .Release.Name }}
+                  key: auth0_client_secret
+            - name: COOKIE_SECRET
+              valueFrom:
+                secretKeyRef:
+                  name: {{ .Release.Name }}
+                  key: cookie_secret
+            - name: NICKNAME_RE
+              value: {{ .Values.Username }}
```
xoen: @s-block we should add `^` and `$` to delimit the beginning and the end of the string here, e.g.:

```yaml
- name: NICKNAME_RE
   value: "^{{ .Values.Username }}$"
```

Basically without those a user called `andyxyz` would be authorised to get into the airflow of another user called `andy` ([based on implementation](https://github.com/ministryofjustice/analytics-platform-auth-proxy/blob/6b64d891f0b2549699664a91713ec3312875ab22/app/base/handlers.js#L118)): https://jsfiddle.net/z4jb58gx/1/

CC: @r4vi

xoen: Ops, looks like `jupyter-lab` is [also affected](https://github.com/ministryofjustice/analytics-platform-helm-charts/blob/master/charts/jupyter-lab/templates/deployment.yaml#L67)...

xoen:
Again, hard to tell from the diff whether there is something wrong with this.
The main thing I've spotted is the nickname regex in the auth proxy which should be more strict, rest looks good.

But obviously you'll have to try to see if Airflow is happy, especially the configuration of it can be fiddly.

charts/airflow-sqlite/templates/webserver-deployment.yml
s-block: I have updated the regex

xoen: 🌟

PR #327, opened 2019-04-10
Rstudio: bump image version

r4vi:
New image startup script, refactored to be more robust against bad conda create of the rstudio
environment.


r4vi:
oops need to update changelog

PR #328, opened 2019-04-16
rbac 0.3.0: `generic-support` users can read namespaces and pods logs

xoen:
- Allow __generic-support__ users to list namespaces. This is to make their
  life easier as they'll find the right namespace without necessarily go
  through the Control Panel. They can already read and list pods so this doesn't
  give support people much more power.
- Allow __generic-support__ users to read pods logs

Ticket: https://trello.com/c/meCj3EJ6


xoen:
Let me know if you can think about any reason why support users shouldn't be able to list namespaces but, most importantly, read pods' logs.

xoen:
Cool thanks to both 🌟 

That's why we need to be sure when we grant permissions to someone.

PR #330, opened 2019-04-29
Bumped Airflow to 1.10.3 in airflow-(k8s|sqlite)

xoen:
- Dockerfile changes: https://github.com/ministryofjustice/analytics-platform-airflow-docker-image/pull/5
- Airflow Changelog: https://github.com/apache/airflow/blob/master/CHANGELOG.txt


xoen:
FYI: Tested this in `dev` (re-release main airflow) and it seemed to work as far as I can tell from a quick inspection. I'm merging this and then upgrading the main airflow in `alpha` as well.

PR #332, opened 2019-05-01
Update airflow-sqlite for use with control panel

s-block:


s-block:
Changes to airflow-sqlite:
- Make airflow-sqlite run in one pod so it can be started/restarted from the control panel
Change to cpanel:
- Add secrets for airflow-sqlite auth proxy and cookie so we can deploy user-airflow

charts/airflow-sqlite/templates/ingress.yml
```diff
@@ -1,11 +1,12 @@
 apiVersion: extensions/v1beta1
 kind: Ingress
 metadata:
-  name: {{ .Release.Name }}
+  name: {{ template "fullname" . }}
```
xoen: We used to use "fullname" initially but it makes the resources name really bad/hard to type 💩, that's why we're now just using release name (which should be unique anyway right?) which is predictable (e.g. if you install the release as `airflow` you know you'll find an Ingress resource called `airflow`).

Any reason why this changed to `{{ template "fullname" . }}`? Maybe it's necessary for some reason I'm missing but if not, it may be nice to just to use the release name. Not a big deal of course :)

xoen: Actually you're right @s-block. In the case of a user tool we probably want to retain the username in the resource name so ignore me 👍

PR #333, opened 2019-05-02
pod-cleaner 0.0.1: Initial release

xoen:
CronJob which deletes old pods in "Succeeded" and "Failed" phase after a
certain number of days.

See kubernetes documentation on phases: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase

Ticket: https://trello.com/c/i1Nxclu3/197-airflow-remove-superseded-pods

### DIFF
--- a/charts/pod-cleaner/Chart.yaml
+++ b/charts/pod-cleaner/Chart.yaml
@@ -1,0 +1,4 @@
+name: pod-cleaner
+description: Delete old completed pods
+version: 0.0.1
+appVersion: 0.0.1
--- a/charts/pod-cleaner/README.md
+++ b/charts/pod-cleaner/README.md
@@ -1,0 +1,50 @@
+# Pod Cleaner
+
+Deletes old pods in `Succeeded`/`Failed` phase:
+
+> Succeeded: All Containers in the Pod have terminated in success, and will not be restarted.
+
+> Failed: All Containers in the Pod have terminated, and at least one Container has terminated in failure. That is, the Container either exited with non-zero status or was terminated by the system.
+
+See kubernetes documentation on [pods phases](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase).
+
+
+## Installing the chart
+
+```bash
+$ helm upgrade --dry-run --debug --install pod-cleaner charts/pod-cleaner --namespace airflow
+```
+
+Install the helm chart in the namespace where you want the cron job
+to delete old pods.
+
+**NOTE**: Remove `--dry-run` and adjust the values file path.
+
+
+## Configuration
+
+| Parameter | Description | Default |
+| --------- | ----------- | ------- |
+| `deleteFailedAfter` | Number of days after which a pod in `Failed` phase should be deleted | `10` |
+| `deleteSucceededAfter` | Number of days after which a pod in `Succeeded` phase should be deleted | `3` |
+| `schedule` | When to run the job that delete the old pods | `"0 5 * * *"` - every day at 5am, see https://kubernetes.io/docs/user-guide/cron-jobs/#schedule |
+
+
+## Caviats
+Pods which are failing but that have `RestartPolicy=Always` will not be
+deleted because from kubernetes perspective these are still in `Running`
+phase.
+
+
+## Technical details
+The cron job will start two containers which will run the `pod_cleaner.sh`
+script. One script will delete the old pods in `Succeeded` phase, the other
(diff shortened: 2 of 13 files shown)


xoen:
@alkar feel free to have a look at this as well if you want.

charts/pod-cleaner/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: 0.0.1
+appVersion: 0.0.1
```
r4vi: why twice?

charts/pod-cleaner/README.md
```diff
@@ -0,0 +1,50 @@
...
+## Configuration
+
+| Parameter | Description | Default |
+| --------- | ----------- | ------- |
+| `deleteFailedAfter` | Number of days after which a pod in `Failed` phase should be deleted | `10` |
+| `deleteSucceededAfter` | Number of days after which a pod in `Succeeded` phase should be deleted | `3` |
+| `schedule` | When to run the job that delete the old pods | `"0 5 * * *"` - every day at 5am, see https://kubernetes.io/docs/user-guide/cron-jobs/#schedule |
+
+
+## Caviats
```
r4vi: Caveats

charts/pod-cleaner/templates/role.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ .Release.Name }}
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - "pods"
+    verbs:
```
r4vi: so this allows service account in airflow ns to delete / list pods?

charts/pod-cleaner/Chart.yaml
xoen: Better be safe than sorry 🤣?  No idea, it must have slipped (or there is a bug in `helm create`, but probably me).

charts/pod-cleaner/README.md
xoen: 🤦🏽‍♀️

xoen: Corrected, thanks ☺️

charts/pod-cleaner/templates/role.yaml
xoen: So the role is created in the namespace where you install the helm release.

Then the `RoleBinding` (note: not `ClusterRoleBinding`) associates this role to the `pod-cleaner` `ServiceAccount`.

The `RoleBinding` is in the `airflow` namespace (or whatever) and therefore grant the service account permission to perform `list` and `delete` on pods in that namespace.

Hope it's clear but if you think this is wrong we can double check.

r4vi: I think it's fine, wonder if that service account already has the permissions to do this (since airflow needs to list and delete pods)

xoen: Relevant documentation: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#rolebinding-and-clusterrolebinding

> A RoleBinding may reference a Role in the same namespace. The following RoleBinding grants the “pod-reader” role to the user “jane” within the “default” namespace. 

```yaml
# This role binding allows "jane" to read pods in the "default" namespace.
kind: RoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: read-pods
  namespace: default
subjects:
- kind: User
  name: jane # Name is case sensitive
  apiGroup: rbac.authorization.k8s.io
roleRef:
  kind: Role #this must be Role or ClusterRole
  name: pod-reader # this must match the name of the Role or ClusterRole you wish to bind to
  apiGroup: rbac.authorization.k8s.io
```

I find a bit sneaky that the "A RoleBinding may reference a Role in the same namespace" which means that the namespace in the `RoleBinding`'s metadata is actually used to determine in which namespace the permissions will be valid.

xoen: Just for the record, as discussed offline.

I don't think re-use service accounts is a best practice as it may lead to running something which only needs {P1, P2} permissions to run as something with more permissions {P1, P2, ...Pn} even if it doesn't need it.

Also, this `pod-cleaner` CronJob should be agnostic of what pods it's cleaner, in our case it's old Airflow tasks pods but it shouldn't really matter.

xoen:
(rebased)

PR #336, opened 2019-05-03
pod-cleaner 0.1.0: Also delete `Pending` pods

xoen:
When a `Pod` fails to start because for example the image cannot be pulled
the `Pod` will be in `Pending` phase.
It makes sense to delete old pods which are failing to start.

### DIFF
--- a/charts/pod-cleaner/CHANGELOG.md
+++ b/charts/pod-cleaner/CHANGELOG.md
@@ -5,6 +5,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
-## [0.0.1] - 2019-03-30
+## [0.1.0] - 2019-05-03
+### Added
+Also delete pods in `Pending` phase.
+
+
+## [0.0.1] - 2019-04-30
 ### Added
 Initial release, yay!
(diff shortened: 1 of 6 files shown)


charts/pod-cleaner/templates/cronjob.yaml
```diff
@@ -19,27 +19,39 @@ spec:
         spec:
           serviceAccountName: {{ .Release.Name }}
           containers:
-          - name: succeeded-pods-cleaner
+          - name: failed-pods-cleaner
```
xoen: FYI: I've just alphabetically sorted the containers here.

xoen:
FYI: I know there is some duplication in the `CronJob` resource but I decided to not overcomplicate things by adding loops, etc in there...

charts/pod-cleaner/CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
-## [0.0.1] - 2019-03-30
+## [0.1.0] - 2019-05-03
+### Added
+Also delete pods in `Pending` phase.
+
+
+## [0.0.1] - 2019-04-30
```
xoen: NOTE: I've corrected the month in the version `0.0.1` as it was wrong 🦊

xoen:
FYI: Released in `dev`/`alpha`.

PR #338, opened 2019-05-03
Webapp: Bump Auth Proxy Version

r4vi:
New auth proxy version can now tell the webapp the email address of the logged
in user via the `USER_EMAIL` header.

This header will not be set unless authentication is enabled


davidread:
Release: https://github.com/ministryofjustice/analytics-platform-auth-proxy/releases/tag/v2.2.0

PR #341, opened 2019-05-13
FluentD Update Repository

djsd123:
`gcr.io/google-containers/fluentd-elasticsearch` has been deprecated in
favour of `gcr.io/fluentd-elasticsearch/fluentd`.

Application versions remain unchanged

- Changed image repo from `gcr.io/google-containers/fluentd-elasticsearch` to `gcr.io/fluentd-elasticsearch/fluentd`
- Changed __fluentd__ image tag from `v2.4.0` to `v2.5.0`

### DIFF
--- a/charts/webapp/Chart.yaml
+++ b/charts/webapp/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
 description: webapp Helm chart for Kubernetes
 name: webapp
-fluentdVersion: 2.5.0
-version: 1.3.25
+fluentdVersion: 1.4.0
+version: 1.3.26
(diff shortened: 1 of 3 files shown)


charts/webapp/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
 description: webapp Helm chart for Kubernetes
 name: webapp
-fluentdVersion: 2.5.0
-version: 1.3.25
+fluentdVersion: 1.4.0
```
xoen: Isn't fluentd still at version `2.5.0` 😕 ?

djsd123: No, I had it wrong in the last PR.  I checked the [Gemfile](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/fluentd-elasticsearch/fluentd-es-image/Gemfile) from a previous commit to verify.

PR #342, opened 2019-05-15
Bump Auth Proxies for Jupyter / Rstudio / Webapp charts

r4vi:
This commit updates the Jupyter / Rstudio / Webapp charts to use the latest
versions of the proxy they use.
In Rstudio's case this is v2.0.0 of rstudio-auth-proxy, and for the others it is
`v3.0.0` of the generic auth proxy


r4vi:
adding `dontmerge` label as I'm considering if I should add the explicit check for being logged in into the generic `auth-proxy` the same way as we do for `rstudio-auth-proxy`.
Jupyter and Webapp currently do the right thing when it comes to websocket authentication but if somebody was to deploy a `webapp` that didn't do the right thing then it's better to be safe than sorry

PR #346, opened 2019-05-22
Bumped auth-proxy tag in airflow-sqlite/jupyter-lab

xoen:
Bumped `auth-proxy` to version [`v4.0.1`](https://github.com/ministryofjustice/analytics-platform-auth-proxy/releases/tag/v4.0.1).

This version removes support for `NICKNAME_RE` and replaces it with
a simple equality check with the `USER` environment variable.

### DIFF
--- a/charts/jupyter-lab/templates/deployment.yaml
+++ b/charts/jupyter-lab/templates/deployment.yaml
@@ -63,8 +63,8 @@ spec:
                 secretKeyRef:
                   name: {{ template "fullname" . }}
                   key: cookie_secret
-            - name: NICKNAME_RE
-              value: "^{{ .Values.Username }}$"
+            - name: USER
+              value: "{{ .Values.Username }}"
             - name: COOKIE_MAXAGE
               value: {{ .Values.authProxy.cookie_maxage | quote }}
             - name: TUNNEL_ENABLED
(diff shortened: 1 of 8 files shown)


xoen:
I think the bump of the auth-proxy for the `airflow-sqlite` helm chart from `v2.1.0` to `v4.0.1` shouldn't break anything a part from `v4.x.x` removing support for `NICKNAME_RE`, am I wrong?

charts/jupyter-lab/templates/deployment.yaml
```diff
@@ -63,8 +63,8 @@ spec:
                 secretKeyRef:
                   name: {{ template "fullname" . }}
                   key: cookie_secret
-            - name: NICKNAME_RE
-              value: "^{{ .Values.Username }}$"
+            - name: USER
```
r4vi: can you try this on your own deployment. the USER on jupyter is usually `joyvan` and I don't know if this will blow a bunch of stuff up if you change it

xoen: Oh, that would be annoying...I wonder if this is a good reason to rename this environment variable altogether to something more explicit and less popular, e.g. like `AUTHORISED_USER` (or `AUTHORIZED_USER` or even `AUTHZ_USER` which to be honest is weird).

Thanks, I'll try.

r4vi: well this is just in the auth-proxy image so actually it should be fine?

xoen: Oh yeah, good point. Even better 💃

PR #347, opened 2019-05-29
webapp 2.0.0: Replaced fluentd with fluent-bit

xoen:
Replaced fluentd with [fluent-bit].

Fluent-bit is a rewrite of fluentd which is much lighter than fluentd (fluend is written in Ruby while fluent-bit is written in C).
This means the logging sidecar container doesn't require much CPU/memory anymore.

Just for comparison the memory limit was `1000Mi` while now it's at `50Mi`.
Even that is conservatively "high" as it's not clear at this stage
how much resources fluent-bit will use in real-life scenarios.

We bumped the major version of the helm chart (to `2.0.0`) as the fluent-bit
resources values are now under `FluentBit.Resources` while previously they
were at the top level.

Also:
- updated config of grep exclude health check requests by agent
  `kube-probe/*` which I can see in Kibana's `agent` record (we used to filter out `Go-http-client` which is
  more brittle).
- added `host` label to webapp's Pods, ConfigMap, Secret, and ServiceAccount to improve filterability of resources

Ticket: https://trello.com/c/vee2XQPW/222-webapps-investigate-possible-switch-to-fluent-bit

[fluent-bit]: https://docs.fluentbit.io/manual/v/1.1/


charts/webapp/templates/deployment.yaml
```diff
@@ -137,39 +138,47 @@ spec:
...
+              subPath: fluent-bit.conf
+            - name: fluent-bit-files
+              mountPath: /fluent-bit/etc/custom_parsers.conf
+              subPath: custom_parsers.conf
+            - name: fluent-bit-files
+              mountPath: /fluent-bit/etc/add_time_nano.lua
+              subPath: add_time_nano.lua
           resources:
-{{ toYaml .Values.resources | indent 12 }}
+{{ toYaml .Values.FluentBit.Resources | indent 12 }}
```
xoen: This is kinda breaking but we don't override the default values in the config repo.
I've bumped the major version anyway.

charts/webapp/values.yaml
```diff
@@ -24,22 +24,23 @@ AuthProxy:
...
+FluentBit:
+  LogLevel: "info"
+  # https://hub.docker.com/r/fluent/fluent-bit
   Image:
-    Repository: gcr.io/fluentd-elasticsearch/fluentd
-    Tag: v2.5.0
+    Repository: fluent/fluent-bit
+    Tag: 1.1.1 # Add `-debug` for debugging image
     PullPolicy: "IfNotPresent"
+  Resources:
```
djsd123: I don't know what `fluent-bit` needs but at the very least I'd increase the `limits`.  The `requests` are very low too.  

requests: cpu: 1 millicore = `0.001` of a core (if I'm not mistaken). Less than 1%

The reason is that your lua function may need more cpu when it runs beside a chatty app.

charts/webapp/templates/fluent-bit-configmap.yml
```diff
@@ -0,0 +1,101 @@
...
+          Name           nginx-error-1st-line
+          Format         regex
+          Regex          ^\d{4}/\d{2}/\d{2} \d{2}:\d{2}:\d{2} \[\w+\] (?<pid>\d+).(?<tid>\d+):
+
+      [PARSER]
+          Name           nginx-error-nth-line
+          Format         regex
+          Regex          ^(?<time>\d{4}/\d{2}/\d{2} \d{2}:\d{2}:\d{2}) \[(?<log_level>\w+)\] (?<pid>\d+).(?<tid>\d+): (?<message>.*)
+
+  add_time_nano.lua: |
```
djsd123: 🥇

charts/webapp/templates/fluent-bit-configmap.yml
```diff
@@ -0,0 +1,101 @@
...
+          DB                  /var/db/nginx-error-log.state.db
+          Multiline           On
+          Parser_Firstline    nginx-error-1st-line
+          Parser_1            nginx-error-nth-line
+
+      # Ignore k8s healthcheck requests
+      [FILTER]
+          Match      web_app.nginx.access
+          Name       grep
+          Exclude    agent ^kube-probe\/.*$
```
djsd123: ⭐️

charts/webapp/templates/fluent-bit-configmap.yml
```diff
@@ -0,0 +1,101 @@
...
+      [FILTER]
+          Match    *
+          Name     lua
+          Script   add_time_nano.lua
+          Call     add_time_nano
+
+      [OUTPUT]
+          Match              *
+          Name               es
+          Type               time
```
djsd123: 👍 Good stuff.  If this were anything else, the whole thing would break.  Needs to be Type `doc` when we upgrade to elasticsearch 7 though.

charts/webapp/values.yaml
xoen: I see what you mean. I agree with you it all depends on the actual usage and logs but as far as I can tell this thing is TINY.

As discussed offline I'm increasing this a little bit to `100m/128Mi` even if I imagine it's still large 😆 .

Bear in mind that [`fluent-bit` helm chart](https://github.com/helm/charts/blob/master/stable/fluent-bit/values.yaml#L174) seems to suggest that `100m/128Mi` are a limit for shipping logs for a k8s node 😱

PR #348, opened 2019-05-31
Init User Trident

djsd123:
* Modified `PersistentVolumeClaim` to provision NFS home on NetApp ONTAP via [Trident](https://github.com/NetApp/trident)
* Modified `ResourceQuota` to allow up to two `PersistentVolumeClaim`'s in a user's namespace
* Modified `ResourceQuota` to allow an aggregate of `20GB` in a user's namespace
* Removed un-required `PersistentVolume` resource.  `PersistentVolume`'s will be auto provisioned with [Trident](https://github.com/NetApp/trident)


charts/init-user/templates/resource-quota.yml
```diff
@@ -5,6 +5,6 @@ metadata:
   namespace: user-{{ .Values.Username }}
 spec:
   hard:
-    persistentvolumeclaims: "1"
-    requests.storage: "15Gi"
-# Allows only 1 volume claim of up to 15GB in user's namespace
+    persistentvolumeclaims: "2"
```
xoen: Is this so that we can keep the old volume claim and allow them to exist together for the migration?

charts/init-user/templates/nfs-persistent-volume-claim.yml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: nfs-home-trident
+  namespace: user-{{ .Values.Username }}
+  labels:
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    app: "trident.netapp.io"
+  annotations:
+    volume.beta.kubernetes.io/storage-class: netapp-file-redundant
```
xoen: This needs to be supported/configured in the cluster right? Maybe I missed an helm chart change/PR, sorry for the stupid question.

charts/init-user/templates/resource-quota.yml
djsd123: Yep

charts/init-user/templates/nfs-persistent-volume-claim.yml
djsd123: Correct.  It's a trident annotation.

r4vi:
⚰

xoen:
🔱

PR #350, opened 2019-06-07
network-policies 0.2.0: Restrict traffic to apps' pods

xoen:
Initial release of helm chart with `NetworkPolicy` resources.
This currently contains only policies to restrict traffic between
webapps' pods but we may add more network policies later (that's why
in the README the installation command install this helm chart
in `default`).

Once this is installed a pod in the `apps-prod` namespace (configurable)
would not be able to talk with other pods in that namespace.

Only traffic allowed will be traffic from the Ingress controller's pod
in the `default` namespace (again, configurable in case these will
run somewhere).

Part of ticket: https://trello.com/c/z0H0xygD/227-prevents-webapp-to-be-able-to-talk-with-each-other

### DIFF
--- a/charts/network-policies/templates/network-policies.yaml
+++ b/charts/network-policies/templates/network-policies.yaml
@@ -17,16 +17,14 @@ metadata:
 spec:
   podSelector: {}
   policyTypes:
-  - Ingress
+    - Ingress
   ingress:
-  - from:
-    - namespaceSelector:
-        matchLabels:
-          name: {{ .Values.ingressController.namespace }}
-{{- if .Values.TEMPORARYnetworkPoliciesWithNamespaceAndPodSelectorsSupported }}
-      podSelector:
-        matchLabels:
-{{ toYaml .Values.ingressController.podsMatchLabels | indent 10 -}}
-{{- end }}
-    ports:
-{{ toYaml .Values.webapps.ports | indent 6 }}
+    - from:
+        - namespaceSelector:
+            matchLabels:
+              name: {{ .Values.ingressController.namespace }}
+          podSelector:
+            matchLabels:
+{{ toYaml .Values.ingressController.podsMatchLabels | indent 14 }}
+      ports:
+{{ toYaml .Values.webapps.ports | indent 8 }}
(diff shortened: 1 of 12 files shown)


xoen:
FYI @djsd123 Once we've upgraded Calico in all clusters we can get rid of `TEMPORARYnetworkPoliciesWithNamespaceAndPodSelectorsSupported` and just always add the `podSelector` bit to the `NetworkPolicy`.

charts/network-policies/templates/network-policies.yaml
```diff
@@ -0,0 +1,32 @@
...
+  podSelector: {}
+  policyTypes:
+  - Ingress
+  ingress:
+  - from:
+    - namespaceSelector:
+        matchLabels:
+          name: {{ .Values.ingressController.namespace }}
+{{- if .Values.TEMPORARYnetworkPoliciesWithNamespaceAndPodSelectorsSupported }}
+      podSelector:
```
xoen: Always add this would break in `alpha` until we upgrade Calico (works fine in `dev` as Mikael upgraded that cluster 🌟)

r4vi: wouldn't it be simpler to just merge this next week and remove this from helm chart?

xoen: Possibly, but I thought we have this option so that we can release on Monday in `alpha` if we want to. And then once Calico is upgraded there we get rid of this.

xoen:
Could you please have another look and let me know what you think? I've mainly [added some helm tests](https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/350/commits/d4c77a2f1a0623685ac5520f2054741cd724dc9c) - The idea of helm tests is interesting but the cleanup doesn't seem to work properly - at least they can be run :).

I've released this in `alpha` this morning and the network policy seem to work (traffic from the outside still works and traffic from another pod times out)

charts/network-policies/templates/network-policies.yaml
xoen: FYI: @djsd123 mentioned he's upgrading `alpha` cluster soon so once that's done I'll get rid of this option and always restrict the pods with `namespaceSelector`+`podSelector`.

PR #351, opened 2019-06-07
Updated chart for Control Panel with merged frontend

andyhd:


charts/cpanel/templates/job.yaml
```diff
@@ -11,27 +11,23 @@ spec:
...
             - name: DB_PASSWORD
-              valueFrom:
-                secretKeyRef:
-                  {{ if .Values.tags.branch -}}
-                  name: "{{ printf "%s-%s" .Release.Name "postgresql" | trunc 63 | trimSuffix "-" }}"
-                  {{- else -}}
-                  name: "{{ template "fullname" . }}"
-                  {{- end }}
-                  key: postgres-password
+              valueFrom: {{ template "postgresPassword" }}
```
djsd123: missing `.`

charts/cpanel/templates/deployment.yaml
```diff
@@ -1,177 +1,57 @@
...
           ports:
             - name: http
-              containerPort: 8000
+              containerPort: {{ .Values.django.port }}
           env:
+          {{- range $key, $value := .Values.env }}
+            - name: {{ $key }}
+              value: "{{ $value }}"
+          {{- end }}
+          {{- range $key, $value := $.Values.secretEnv }}
```
djsd123: `$.Values.secretEnv` is not in scope?

charts/cpanel/templates/deployment.yaml
```diff
@@ -1,177 +1,57 @@
...
-    app: "{{ template "fullname" .}}"
+    app: "{{ .Chart.Name }}"
+    host: {{ template "host" . }}
 spec:
   replicas: 3
   template:
     metadata:
       labels:
-        app: "{{ template "fullname" . }}"
+        app: "{{ .Chart.Name }}"
```
xoen: I'd add the `host` label to the pods as well as it's convenient when filtering, etc...

charts/cpanel/templates/cluster-role.yaml
```diff
@@ -1,24 +1,16 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: {{ template "fullname" . }}
+  name: "{{ .Chart.Name }}"
```
xoen: In the other helm charts (for tools) I added the `host` label to pretty much all resources. Maybe it's overkill but at the same time it could be useful to have this convention for migrations, etc...

andyhd: Ok, I'll make sure its added to everything 👍

charts/cpanel/templates/_helpers.tpl
```diff
@@ -1,23 +1,38 @@
...
+    {{ template "postgresRelease" . }}
+  {{- else -}}
+    "{{ .Values.postgresql.postgresHost }}"
+  {{- end -}}
+{{- end -}}
+
+{{/*
+Postgres password
+*/}}
+{{- define "postgresPassword" -}}
```
xoen: I'm not sure I understand the need for the complexity.
This used to be in the template, This is generating the `secretKeyRef` when what changes is really only the secret name, so if this was to generate only that it could be called something like `secretName`.

I guess this is complicated because we pull the password out of the postgres helm chart (if I understand correctly), I wonder if we could set this password instead and use the same secret and get rid of the if altogether.

andyhd: The password comes out of the postgres helm chart *if* it is a per-branch deployment, otherwise the secret is supplied by config.  I factored out the whole `secretKeyRef` because it is is two places (the deployment and the job), but maybe that confuses things too much.

xoen: You don't have to change it if you think it's better. It was just a thought.

PR #353, opened 2019-06-25
unidler v4.0.1: Use strategic merge patch instead of JSONPatch

xoen:
Use new version of the unidler which uses strategic merge patch instead of
JSONPatch.

Strategic merge is more declarative and more robust. When deleting
a key which is not there it would consider that part of the patch as already
applied instead of raising an error as in the case of JSONpatch.

- [unidler PR](https://github.com/ministryofjustice/analytics-platform-go-unidler/pull/10)
- Part of ticket: https://trello.com/c/vr4LPcde/241-investigate-fix-idling-unidling-problems-caused-by-k8s-cluster-upgrade-new-version-of-k8s

### DIFF
--- a/charts/unidler/values.yaml
+++ b/charts/unidler/values.yaml
@@ -1,7 +1,7 @@
 # Docker image
 image:
   name: quay.io/mojanalytics/go-unidler
-  tag: v1.0.1
+  tag: v1.0.2
   pullPolicy: IfNotPresent
 
 replicaCount: 3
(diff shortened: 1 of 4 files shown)


charts/unidler/values.yaml
```diff
@@ -1,7 +1,7 @@
 # Docker image
 image:
   name: quay.io/mojanalytics/go-unidler
-  tag: v1.0.0
+  tag: v1.0.2
```
xoen: Had to create another release to fix a minor build problem: https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/353/commits/edcd8834ff5c894f5731b4076da4b816071f988f

PR #356, opened 2019-07-04
Add nginx reverse proxy to Control Panel for static files

andyhd:


charts/cpanel/README.md
```diff
@@ -12,9 +12,9 @@ To install the chart:
 ```bash
 helm install mojanalytics/cpanel \
   --name cpanel-$BRANCH_NAME \
-  --set API.Environment.DEBUG=True \
-  --set API.Image.Tag=$DOCKER_IMAGE_TAG \
-  --set API.Branch=$BRANCH_NAME \
+  --set env.DEBUG=True \
```
xoen: Always good to keep the README up-to-date 🌟 

Thinking aloud: I've also started to replace the `helm install` commands with `helm upgrade --install` (I also add `--dry-run --debug`). Most of the time we really upgrade releases - having to fiddle to convert the command is error-prone, especially if someone didn't release a chart in awhile (or ever!).

charts/cpanel/README.md
```diff
@@ -38,34 +38,33 @@ In Auth0 you need to install Extension 'Auth0 Authorization':
...
-| `API.Environment.ELASTICSEARCH_PORT` | Elasticsearch port number | `9243` |
-| `API.Environment.ELASTICSEARCH_USERNAME` | Elasticsearch username | `""` |
-| `API.Environment.ELASTICSEARCH_PASSWORD` | Elasticsearch password | `""` |
-| `AWS.DefaultRegion` | AWS region | `eu-west-1` |
-| `AWS.IAMRole` | IAM role assumed by the instance running the API | |
+| `aws.defaultRegion` | AWS region | `eu-west-1` |
+| `aws.iamRole` | IAM role assumed by the instance running the API | |
+| `env.DEBUG` | Used to set Django DEBUG mode | `False` |
+| `env.ELASTICSEARCH_PORT` | Elasticsearch port number | `9243` |
+| `env.ENABLE_K8S_RBAC` | Set to `True` if k8s RBAC is enabled | `False` |
```
xoen: I didn't realise this defaulted to `False` 😆- we've been using RBAC for ages now. At some point _in the future_ we may want to change the default to `True` and remove the need to override the value.

charts/cpanel/templates/_helpers.tpl
```diff
@@ -16,23 +16,28 @@ Old Control Panel API hostname
 Postgres release
 */}}
 {{- define "postgresRelease" -}}
-"{{ printf "%s-%s" .Release.Name "postgresql" | trunc 63 | trimSuffix "-" }}"
+{{ printf "%s-%s" .Release.Name "postgresql" | trunc 63 | trimSuffix "-" }}
```
xoen: Stupid question: We don't need to quote this?

charts/cpanel/templates/_helpers.tpl
```diff
@@ -16,23 +16,28 @@ Old Control Panel API hostname
...
 {{- end -}}
 
 {{/*
 Postgres hostname
 */}}
 {{- define "postgresHost" -}}
   {{- if .Values.tags.branch -}}
-    {{ template "postgresRelease" . }}
+    {{ include "postgresRelease" . }}
```
xoen: I always have to try these templates so maybe this is fine but do we need `| quote`? Is this changing because it was broken?

charts/cpanel/templates/deployment.yaml
```diff
@@ -17,7 +17,18 @@ spec:
     spec:
       serviceAccountName: "{{ .Chart.Name }}"
       containers:
-        - name: "{{ .Chart.Name }}"
+        - name: reverse-proxy
+          image: "nginx:1.16.0-alpine"
```
xoen: Ideally we should aim at running all containers as non-root (unless strictly necessary).

You may want to try to use a Docker image which doesn't run as root, [we did something similar in the `webapp-template` Docker image](https://github.com/moj-analytical-services/webapp-template/blob/master/Dockerfile) so maybe it's a good idea to do it here too.

It would kinda future-proof this for when/if we'll decide to restrict root-containers.

charts/cpanel/templates/deployment.yaml
```diff
@@ -17,7 +17,18 @@ spec:
...
+          imagePullPolicy: IfNotPresent
+          ports:
+            - name: http
+              containerPort: 80
+          volumeMounts:
+            - name: nginxconf
+              mountPath: /etc/nginx/conf.d
+            - name: staticfiles
+              mountPath: /usr/share/nginx/html
+        - name: django
```
xoen: Not essential or necessary to change but I'm wonder if we should just call this container `backend` (if I understand correctly). And maybe the other one something like `staticfiles` or even `nginx`. Again, assuming I understand the role of these containers correctly. It may help self-document things up.

charts/cpanel/templates/env-configmap.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: env-configmap
```
xoen: This could potentially cause problems/confusion down the line. Assuming we still install the CP helm chart(s?) in the `default` namespace and if we use a similar pattern in another helm chart we may end up with trying to create two different `ConfigMap`s with the same name in the same namespace or worst, succeed/break something accidentally.

It's a good argument for using the release name for the resources.

Speaking of naming: I always have two think/grep resources to figure out the pods/resources belonging to the CP. I know that now the CP URL used by the users has the release name (which is probably not ideal) so maybe we shouldn't break that, unless we plan around it...

...but maybe we could at some point change the release name for the control panel to just  `control-panel` for predictability. If we still want to keep frontend/backend separated we may want to call the API `control-panel-api` and the frontend `control-panel-frontend` or just `control-panel`. The `master` bit could be omitted when releasing the main release "production" chart.

Sorry, just thinking aloud.

charts/cpanel/templates/secrets.yaml
```diff
@@ -9,6 +9,21 @@ metadata:
 type: Opaque
 data:
   postgres-password: {{ .Values.postgresql.postgresPassword | b64enc | quote }}
-  {{- range $key, $value := .Values.secretEnv }}
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: env-secrets
```
xoen: Again, see other comment on having a k8s resource with a fairly generic name like this.

Bear in mind that if we have multiple releases of this chart (e.g. the main one + other for branches) I think this may blow up the upgrade/install or worst...succeed with weird or even dangerous consequences (e.g. branch-release talking with production DB, etc...)

charts/cpanel/templates/service.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
 spec:
   ports:
   - port: 80
-    targetPort: {{ .Values.django.port }}
+    targetPort: 80
```
xoen: Is there any way to not run the CP on port `80` which requires the container to use  `root` user? Also, there is some value in having that in a value (even if the default one will be used 99.9% of the time), as for example it made my life easier in a couple of helm charts when I was changing the image to not run as `root` so you may keep the value.

xoen: Also, I see we still use `.Values.django.port` in the rest of the helm chart so this may introduce bugs.

xoen:
@andyhd looks generally good 👍, I left some of my usually annoying comments.
Some of them are just thoughts but you may want to make some tweaks (or not).

charts/cpanel/templates/service.yaml
andyhd: `.Values.django.port` is 8000 - The Django app listens on port 8000 and Nginx listens on port 80 and proxies to port 8000.

charts/cpanel/templates/env-configmap.yaml
andyhd: Yep, I will make the name `{{ .Chart.Name }}-env-`

charts/cpanel/templates/secrets.yaml
andyhd: 👍 I'll change it to `{{ .Chart.Name }}-env-secrets`

charts/cpanel/templates/deployment.yaml
andyhd: I agree on changing it from `django` to something that describes the role, maybe `backend` - but `staticfiles` is not any better than `reverse-proxy` because it does both reverse proxy and serve static files. `nginx` is a bit too vague.

charts/cpanel/templates/deployment.yaml
andyhd: That would defeat the purpose of using the official image. I could have made a custom Docker image which runs as non-root and contains the static files, but then we'd need a bunch of extra build steps and a new repository.

andyhd: I will switch to [`docker-nginx-unprivileged`](https://github.com/nginxinc/docker-nginx-unprivileged)

charts/cpanel/templates/_helpers.tpl
andyhd: These helpers need to return an unquoted string - the values are quoted at the calling side.

charts/cpanel/templates/_helpers.tpl
andyhd: No, in fact, I need the value to be unquoted or else things break.

andyhd: Specifically, the value returned by calling `{{ include "postgresRelease" . }}` is then `b64encode`d, and then `quote`d - so if it's already quoted, the quotes get base64 encoded too :(

charts/cpanel/templates/deployment.yaml
xoen: That image is the one I used in the other repo and that I was suggesting 👍

charts/cpanel/templates/service.yaml
xoen: Of course! now all traffic goes from service to nginx container! Still, I think there is value into having this in a value instead of hardcoding, for a start, it would self-document it and avoid confusion.

E.g. if I saw `{{ .Values.django.port }}` replaced with something like `{{ .Values.proxy.port }}` I would have been less confused :)

charts/cpanel/templates/deployment.yaml
xoen: Speaking of values, I know values are not really constants (in fact the opposite) but it may be a good idea to move the image name/tag into a value as what we do in other helm charts so it may make things more consistent/predictable with the rest.

charts/cpanel/templates/service.yaml
andyhd: I agree it would be self-documenting, but making it a value would make it `--set`able - if you changed the value, nothing would work.

andyhd:
I've realised a bunch of my changes to the deployment didn't get committed, for some reason. Missing volumes and initContainer. 😨

charts/cpanel/templates/service.yaml
xoen: Well, if you set the value to `9000` it would change in both places (service and container) and continue to work. If use set this to `80` and we restrict `root` containers it would fail but if we hardcode this value and we upgrade the helm chart and miss to update one of the two port values we'd break it.

As said, it was handy to have the port as a value instead of hardcoded in the `webapp` helm charts as it made the migration to non-root containers easier.

But it's your call, I think having the value is just less error prone as you use meaningful names instead of numbers.

But as I said, the changes look good, these are more observations than anything :)

xoen:
@andyhd oh yes, didn't spot that, weird 😱

charts/cpanel/templates/service.yaml
andyhd: It wouldn't change in the nginx config, though

andyhd: I will make it a value, because I agree it makes the chart self-documenting.

PR #357, opened 2019-07-12
Add parameter secrets to helm chart

s-block:


charts/concourse-org-pipeline/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [0.2.4] - 2019-07-12
```
xoen: From a [semantic versioning](https://semver.org/) perspective: If this is adding a feature in a non-breaking way it should probably be version `0.3.0`.

PR #359, opened 2019-07-26
restarter v0.0.1: Initial release

xoen:
Internal API to restart applications in a given
namespace.

This API doesn't have an Ingress resource so it's
only accessible from within the k8s cluster (at `${RELEASE_NAME}.${NAMESPACE}.svc.cluster.local`).

Furthermore, the service account for this API is only able to list and
patch Deployments within the namespace where the chart/API is been
released to.

Part of ticket: https://trello.com/c/zb9HsZca/254-investigate-ways-to-trigger-apps-restart-when-data-changes

### DIFF
--- a/charts/restarter/templates/NOTES.txt
+++ b/charts/restarter/templates/NOTES.txt
@@ -1,0 +1,1 @@
+The internal API will be available at {{ .Release.Name }}.{{ .Release.Namespace }}.svc.cluster.local
--- a/charts/restarter/CHANGELOG.md
+++ b/charts/restarter/CHANGELOG.md
@@ -1,0 +1,10 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+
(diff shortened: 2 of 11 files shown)


xoen:
FYI: Need to do some manual end-to-end testing in `dev` but at least it deploys and it runs...

charts/restarter/templates/NOTES.txt
```diff
@@ -0,0 +1 @@
+The internal API will be available at {{ .Release.Name }}.{{ .Release.Namespace }}.svc.cluster.local
```
r4vi: nice you actually used the NOTES.txt for something useful

xoen: Always trying ☺️. We're users too.

xoen: I always try when it makes sense :) 

Good to have some meaningful information after you just installed an helm chart.

xoen:
FYI: I've added [this `NetworkPolicy`](https://github.com/ministryofjustice/analytics-platform-helm-charts/blob/1dd9e89316c3ce9e79970e376e1146af94050a3d/charts/restarter/templates/networkpolicy.yaml) as if we deploy this in the `apps-prod` namespace only inbound traffic from internet (Ingress Controller pods) will be allowed - which is good - but also a pain :P

PR #360, opened 2019-07-31
Adding exclusions to nfs backup

djsd123:
The below directories seem to be the usual culprits that cause the nfs
backup job to either stall or fail entirely.

I don't think there is much value in backing these directories to `s3` but please shout
if you disagree.

```
.conda
miniconda3-r
.local
Git_Repos
SAT+Shapes
py-safety
.virtualenvs
comparisons
Python
```

Already deployed `(31/07/2019)`

Will check later or tomorrow to see if it had the desired effect.

### DIFF
--- a/charts/nfs-backup/Chart.yaml
+++ b/charts/nfs-backup/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for Kubernetes
 name: nfs-backup
-version: 0.2.0
+version: 0.2.1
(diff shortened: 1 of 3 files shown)


r4vi:
agreed

djsd123:
@andyhd, although I have not seen the files mentioned causing a problem it's a good idea.  I'll see what we've excluded in `. gitignore` and `. dockerignore` from the config-user chart.

djsd123:
Since, the alert, `cronjob=nfs-backup-nfs-backup` has not been triggered and the backup has consistently completed.  Added `node_modules` to list.

PR #364, opened 2019-08-19
Bump Webapp chart: v5.1.0

r4vi:
This changes the response code the auth-proxy sends back when user is not in the
ip_whitelist from 500 to 403


obutterbach:
Looks good to me :)

xoen:
Sorry `403 Forbidden` 🤦‍♂ - coffee is kicking in :)

PR #368, opened 2019-08-23
Bumped auth-proxies to v5.2.1

xoen:
This version of [`auth-proxy`](https://github.com/ministryofjustice/analytics-platform-auth-proxy) removes the non-standard `prompt=none`.
Also note that v5.2.0 introduced support for logic to log into kibana.

Charts updated:
- airflow-sqlite
- jupyter-lab
- rstudio
- webapp

**NOTE**: `kubernetes-dashboard` and `prometheus-resources` use an older
version of the proxy - we should probably update those as well:
- [X] update `kubernetes-dashboard`
- [x] update `prometheus-resources`

See auth-proxy releases for more details:
- https://github.com/ministryofjustice/analytics-platform-auth-proxy/releases/tag/v5.2.1
- https://github.com/ministryofjustice/analytics-platform-auth-proxy/releases/tag/v5.2.0

### DIFF
--- a/charts/airflow-sqlite/CHANGELOG.md
+++ b/charts/airflow-sqlite/CHANGELOG.md
@@ -5,6 +5,17 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+## [0.2.2] - 2019-08-23
+### Changed
+Bumped auth-proxy to version v5.2.1.
+This version removes the non-standard `prompt=none`.
+Also note that v5.2.0 introduced support for logic to log into kibana.
+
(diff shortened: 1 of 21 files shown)


xoen:
I've bumped the `auth-proxy` version on pretty much everything (unless I've missed something).

`kubernetes-dashboard` and `prometheus-resources` used ancient versions but their use is relatively basic so I reckon It Shouldn't Break (TM) 🤣 - in these two cases I've just set empty `USER` to be sure that we don't accidentally trigger the authorisation layer somehow (e.g. the one that check that the logged user's username is equal to that variable).

@djsd123 you know more about these two helm chart so have a look. I'll bump the tag manually on these `Deployment`s in `dev` and see how it goes.

xoen:
Merging this.

PR #369, opened 2019-08-27
Webapp: Add security headers by default

r4vi:
This adds some pretty safe security headers to all responses by default.

This should not be a breaking change unless an app is doing something weird like
embedding itself on another site via an IFRAME

ticket: https://trello.com/c/wpEJaXXt/36-sdt-changes-based-on-zap-result


r4vi:
@xoen we debated where to do this in a few places, but this was the most non-invasive
- didn't want to do it on auth-proxy because more than just webapps use it
- didn't want to do it in the app because setting headers is hard in R plus don't want to replicate for every app

xoen:
@r4vi I quite like this because as you said is not very invasive 👍

PR #370, opened 2019-08-27
Add Redis to cpanel for SSE and background tasks

andyhd:
* Adds a chart dependency on Redis, which is used to handle message passing between the Django app and background tasks and as a channel layer for Server Sent Events


charts/cpanel/values.yaml
```diff
@@ -51,6 +51,10 @@ django:
   port: 8000
   settingsModule: "controlpanel.settings"
 
+redis:
+  persistence:
+    enabled: false
```
r4vi: is this a redis without a password?

r4vi: have a look @xoen's work on `redis` at: https://github.com/ministryofjustice/analytics-platform-helm-charts/blob/master/charts/airflow-k8s/templates/webserver-deployment.yml to see how to at least set a password, and for bonus points a networkpolicy so that only cpanel pods can talk to it

r4vi: this is probably a better file to look at: https://github.com/ministryofjustice/analytics-platform-helm-charts/blob/master/charts/airflow-k8s/values.yaml#L57

andyhd: Yeah, good point - need a password and to lock down access. Also, I've forgotten to add a pod for the background task worker.

PR #371, opened 2019-08-29
Refactor cpanel for scalability

andyhd:
Not sure if this is sensible - please comment!

* Split deployment into 3 parts - app, reverse-proxy and worker - this allows us to scale each component separately
* New service for app and generate reverse-proxy nginx.conf from template
* Split all resources into their own files
* Replaced my mistaken use of `.Chart.Name` with `.Resource.Name` - [Helm best practice](https://helm.sh/docs/chart_best_practices/#resource-naming-in-templates)


andyhd:
I missed the part where I pass the nginx.conf configmap to the reverse-proxy initContainer. Oops!

xoen:
I know it's a WIP but before it slips off my mind, it may be worth rebase and then bump version in `Chart.yaml` and add an entry to `CHANGELOG.md`. I'm having a look at this soon-ish ☺️

charts/cpanel/templates/app-deployment.yaml
```diff
@@ -0,0 +1,47 @@
...
+spec:
+  replicas: 3
+  selector:
+    matchLabels:
+      app: "{{ .Chart.Name }}"
+  template:
+    metadata:
+      labels:
+        app: "{{ .Chart.Name }}"
+        {{ .Release.Name }}-redis-client: true
```
xoen: I've recently started to add the `host` label also to the pods as it makes things easy to maintain, e.g. you can easily find (or delete...) pods by copying the host and using `kubectl get pods -lhost=...`.

Speaking of, maybe we could have a `component` label as well so that we can potentially distinguish these `Pods`/`Deployment`, e.g. `component=app`,`component=worker`, `component=reverse-proxy`, or something like that. It's something similar to what we do [in the `airflow` chart](https://github.com/ministryofjustice/analytics-platform-helm-charts/blob/master/charts/airflow-k8s/templates/webserver-deployment.yml#L16).

charts/cpanel/templates/app-deployment.yaml
```diff
@@ -0,0 +1,47 @@
...
+      annotations:
+        iam.amazonaws.com/role: "{{ .Values.aws.iamRole }}"
+    spec:
+      serviceAccountName: "{{ .Release.Name }}"
+      volumes:
+      - emptyDir: {}
+        name: nginxconf
```
xoen: Does the app needs this `nginxconf` volume?

charts/cpanel/templates/app-deployment.yaml
```diff
@@ -0,0 +1,47 @@
...
+      - emptyDir: {}
+        name: staticfiles
```
xoen: What about this `staticfiles` volume? How is it used? Aren't static files built/kept in the Docker image?

charts/cpanel/templates/app-deployment.yaml
```diff
@@ -0,0 +1,47 @@
...
+      containers:
+        - name: backend
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+          imagePullPolicy: "{{ .Values.image.pullPolicy }}"
+          ports:
+            - name: http
+              containerPort: {{ .Values.django.port }}
+          envFrom:
```
xoen: This `envFrom` trick is pretty neat 😎

charts/cpanel/templates/app-deployment.yaml
```diff
@@ -0,0 +1,47 @@
...
+          - configMapRef:
+              name: "{{ .Release.Name }}-env-configmap"
+          - secretRef:
+              name: "{{ .Release.Name }}-env-secrets"
+          - secretRef:
+              name: "{{ .Release.Name }}-db-env"
+          readinessProbe:
+            httpGet:
+              path: /health
```
xoen: I don't know/remember what's in the app but just for consistency, shouldn't this be `/healthz` like for the others? Just to avoid any confusion/possible problems down the line.

charts/cpanel/templates/app-deployment.yaml
andyhd: Yes - I mentioned that in the first comment on this PR - I'm creating the nginx configmap, but then I forgot to actually use it! 🤦‍♂

charts/cpanel/templates/app-deployment.yaml
andyhd: I was thinking about adding a component label - glad we have a precedent! And yeah, the hosts label should be there too 👍

charts/cpanel/templates/app-deployment.yaml
andyhd: This is already in use. The staticfiles are copied from the Docker image into the nginx container so that it can serve them without proxying to Django (which doesn't serve static files when `DEBUG=False` anyway)

charts/cpanel/templates/app-deployment.yaml
andyhd: What's the reason for the weird `/healthz` path? This path points to django-health-check, which shows a page with health status of various django components

charts/cpanel/templates/app-service.yaml
```diff
@@ -0,0 +1,14 @@
...
+  name: "{{ .Release.Name }}"
+  labels:
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    app: "{{ .Chart.Name }}"
+    host: {{ template "host" . }}
+spec:
+  ports:
+  - port: {{ .Values.django.port }}
+  selector:
+    app: "{{ .Chart.Name }}"
```
xoen: So the main app's `Service` is pointing the pod with `app=$CHART_NAME` while the reverse proxy's `Service` is pointing to the pod with `app=$CHART_NAME-reverse-proxy`, or something like that.

Alternatively all the pods could have the same `app=$CHART_NAME` (shouldn't be `.Release.Name`? That's what usually we do to allow for different releases of the same helm chart in a namespace) and add `component=...` and have selector with both `app` and `component`, similar to what we do in [the`airflow` helm chart](https://github.com/ministryofjustice/analytics-platform-helm-charts/blob/master/charts/airflow-k8s/templates/service.yml#L12):

```yaml
selector:
    app: {{ .Release.Name }}
    component: "app" # or whatever
```

just my 2 pence of course.

charts/cpanel/templates/app-deployment.yaml
xoen: sorry, missed that 🌹

charts/cpanel/templates/app-service.yaml
andyhd: Thanks 👍  - I had a feeling the labels needed work

charts/cpanel/templates/app-deployment.yaml
xoen: It's just what you'll see everywhere in kubernetes. I suppose the reason was predictability, e.g. if I was investigating something I may making a request to `/healthz` without thinking too much. At the same time if editing this helm chart I may think it's a typo and "correct" it causing problems.

But it doesn't really matter from k8s perspective as long as whatever endpoint in the probe exists.

charts/cpanel/templates/app-service.yaml
xoen: I don't know what's the best practice there but that pattern kind of makes sense and we may well use it until we find a better one 🤷‍♂ - again, not that k8s cares about this, it would work anyway.

charts/cpanel/templates/app-deployment.yaml
xoen: It's one of these little consistency things that I'm finding useful while investigating something 😍

charts/cpanel/templates/app-deployment.yaml
xoen: Hang on, I saw the copying in the init container of the reverse proxy deployment but does the app deployment needs this `staticfiles` empty dir volume? I'm probably missing something sorry.

andyhd: Ah, you are right about this and the nginxconf volume - this is copypasta - the app deployment doesn't need either of these volumes. Sorry!

andyhd: These belong to the reverse-proxy deployment

xoen: Pasta 🍝 😍  nomnomnomnom...

davidread:
@xoen recently split the worker off into a separate pod, superseding this PR

PR #376, opened 2019-08-30
New Environment: Cluster Autoscaler

djsd123:
[Trello](https://trello.com/c/WpMxBEOb)

- Changed tag version from `v1.3.5` to `v1.12.3`
- Increased memory limit from `300mi` to `500mi`
- Modified `Clusterrole` by adding `apps` apiGroup to read-only on
`ReplicaSets` and `DaemonSets`
- Removed `env` value as instances reside in separate AWS accounts

**Deployed**

### DIFF
--- a/charts/cluster-autoscaler/templates/deployment.yaml
+++ b/charts/cluster-autoscaler/templates/deployment.yaml
@@ -6,7 +6,7 @@ metadata:
   labels:
     app: cluster-autoscaler
   annotations:
-    iam.amazonaws.com/role: "{{ .Values.env }}-cluster-autoscaler"
+    iam.amazonaws.com/role: "cluster-autoscaler"
 spec:
   replicas: {{ .Values.replicaCount }}
   selector:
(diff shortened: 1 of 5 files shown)


charts/cluster-autoscaler/templates/deployment.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   labels:
     app: cluster-autoscaler
   annotations:
-    iam.amazonaws.com/role: "{{ .Values.env }}-cluster-autoscaler"
+    iam.amazonaws.com/role: "cluster-autoscaler"
```
xoen: Would this be a breaking change? I'm asking as in that case we may want to bump the version of the helm chart from `0.x` to `1.x`?

djsd123: Indeed 👍

djsd123:
Not sure this relevant now

PR #377, opened 2019-09-02
add selector to daemonset spec

sh31:
* Added selector to daemonset spec as required by the K8s apps/v1 api


obutterbach:
Keep me updated if you merge your PR, I would have to re-update mine 😄

PR #379, opened 2019-09-09
Add old cpfrontend host to ingress

andyhd:
This will make the new Control Panel receive requests to `https://cpanel-master.services...`

I'm assuming the way to actually remove the old frontend is to just `helm delete` it, but let me know if you have a nicer way

### DIFF
--- a/charts/cpanel/templates/ingress.yaml
+++ b/charts/cpanel/templates/ingress.yaml
@@ -22,6 +22,12 @@ spec:
       - backend:
           serviceName: "{{ .Chart.Name }}"
           servicePort: 80
+  - host: {{ template "old_cpfrontend_host" . }}
+    http:
+      paths:
+      - backend:
+          serviceName: "{{ .Chart.Name }}"
+          servicePort: 80
   tls:
   - hosts:
     - {{ template "host" . }}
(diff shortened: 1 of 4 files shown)


charts/cpanel/templates/ingress.yaml
```diff
@@ -22,6 +22,12 @@ spec:
       - backend:
           serviceName: "{{ .Chart.Name }}"
           servicePort: 80
+  - host: {{ template "old_cpfrontend_host" . }}
```
xoen: Where is this defined?

xoen:
I thought new CP URL would be `controlpanel.{{ Values.servicesDomain }}`? Maybe I got confused.

andyhd:
Yes it is `controlpanel.{{ .Values.serviceDomain }}` - this is so that the old URL points to the new Control Panel

r4vi:
will links to assets like: https://cpanel-master.services.alpha.mojanalytics.xyz/static/images/gov.uk_logotype_crown.svg
used in https://github.com/ministryofjustice/analytics-platform-auth0/blob/alpha/pages/login.html#L146 continue to work?

andyhd:
Ah, no. Should I make that work or should we get the new URL into the login page template?

r4vi:
@andyhd what's easier? it's pretty easy to update those templates but I wonder where else it's being used

andyhd:
[Technically, we shouldn't be using the crown logo anyway.](https://www.gov.uk/service-manual/design/making-your-service-look-like-govuk) I don't know what else is deep-linking into the Control Panel frontend. Probably just the login page. The new Control Panel uses the GOV.UK Design System, which may well have different image assets from the old GOV.UK Frontend and Elements stuff too.

andyhd:
Probably [one of these](https://www.google.com/search?q=moj+logo&tbm=isch)

r4vi:
this is nice

PR #380, opened 2019-09-10
User Read ClusterRole

djsd123:
Conversation earlier prompted the need for a read-only role on user
namespaces.

- Added __user-read__ `ClusterRole` with intent to be used primarily by the [__init-user__](../init-user) helm chart
and/or any other entity that can make use of it. i.e. `ServiceAccount` binding.
- __user-read__ `ClusterRole` scoped to read `ingresses`, `services`, `deployments`, `pods` and `configmaps`

### DIFF
--- a/charts/init-user/templates/role-binding.yml
+++ b/charts/init-user/templates/role-binding.yml
@@ -27,3 +27,19 @@ subjects:
   - apiGroup: rbac.authorization.k8s.io
     kind: Group
     name: user-support
+
+---
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: user-read
+  namespace: user-{{ .Values.Username }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: user-read
+subjects:
+  - apiGroup: rbac.authorization.k8s.io
+    kind: Group
+    name: user-read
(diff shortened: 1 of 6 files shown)


charts/init-user/templates/role-binding.yml
```diff
@@ -27,3 +27,19 @@ subjects:
   - apiGroup: rbac.authorization.k8s.io
     kind: Group
     name: user-support
+
+---
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
```
xoen: Maybe I'm not understanding what's the use case but...

wouldn't be better to have a `ClusterRole` for reading stuff like (there is a `view` one but its list of permissions is too long for my taste) and then just create a `RoleBinding` in the user namespace that refers to this unique `ClusterRole`? (it's possible right?) so that we avoid having a lot of roles that will get out-of-date for sure in few months.

Maybe I'm missing something.

I think there are some examples in the RBAC documentation: `https://kubernetes.io/docs/reference/access-authn-authz/rbac/#rolebinding-and-clusterrolebinding`

xoen: This is what we do in for the `user-{username}` `RoleBinding` which is binding the `admin` `ClusterRole`.

...but I am an idiot because I just realised that this is exactly what you did sorry 😆 - I just didn't realise from the PR diff that `user-read` was indeed a `ClusterRole` and part of the other helm chart - I shouldn't review PR while my brain is fried sorry 😝

PR #382, opened 2019-09-17
RBAC: Add concourse-deployments ClusterRole

r4vi:
And bind it to the concourse-main default service account so that concourse can
report on rollout status.

https://trello.com/c/ofx66AHU/315-show-deployment-rollout-status-for-concourse-app-builds

### DIFF
--- a/charts/rbac/templates/concourse-deployments.yaml
+++ b/charts/rbac/templates/concourse-deployments.yaml
@@ -1,0 +1,39 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: concourse-deployments
+  labels:
+    chart: {{ template "rbac.chart" . }}
+  namespace: {{ .Values.teams.concourseDeploymentStatus.appsNamespace }}
+  annotations:
+    description: {{ .Values.teams.concourseDeploymentStatus.description | quote }}
+rules:
+  - apiGroups:
+      - "extensions"
+      - "apps"
+    resources:
+      - deployments
+    verbs:
+      - get
+      - watch
+      - list
+
+---
+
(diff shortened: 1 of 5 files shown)


charts/rbac/templates/concourse-deployments.yaml
```diff
@@ -0,0 +1,38 @@
...
+metadata:
+  name: concourse-deployments
+  labels:
+    chart: {{ template "rbac.chart" . }}
+  namespace: {{ .Values.teams.concourseDeploymentStatus.appsNamespace }}
+  annotations:
+    description: {{ .Values.teams.concourseDeploymentStatus.description | quote }}
+rules:
+  - apiGroups:
+      - "extensions"
```
djsd123: Might want to add `apps` too

r4vi: @djsd123 it doesn't need it, so just adding the minimum - what's the reason for adding?

djsd123:
@r4vi We usually add a row to the README table

charts/rbac/templates/concourse-deployments.yaml
djsd123: Just thinking if concourse was ever upgraded in future.  Not a problem though

r4vi:
@djsd123 updated readme and changelog :scroll: good spot

djsd123:
LGTM

charts/rbac/templates/concourse-deployments.yaml
r4vi: I see what you mean now, `deployments` will move to the `apps` api group eventually so good to do it now so things don't break when it does. I'll make the change

PR #383, opened 2019-09-18
S3 proxy

isichei:


charts/webapp/CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [2.2.1] - 
+### Changed
+Added HealthCheck option for deploy.json. Defaults to "/?healthz" for standard webapp and is "/healthz" in deploy.json for s3 proxy webapp deployment.
+
```
r4vi: following lines are a copypasta?

charts/webapp/CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [2.2.1] -
```
r4vi: date?

charts/webapp/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
 description: webapp Helm chart for Kubernetes
 name: webapp
-version: 2.2.0
+version: 2.2.1
```
r4vi: probably should be `2.3.0` because it adds a new feature, rather than `2.2.1` which would be a bugfix

r4vi:
```suggestion
version: 2.3.0
```

r4vi: probably should be 2.3.0 because it adds a new feature, rather than 2.2.1 which would be a bugfix

charts/webapp/templates/deployment.yaml
```diff
@@ -92,7 +92,7 @@ spec:
             timeoutSeconds: 60
           livenessProbe:
             httpGet:
-              path: /?healthz
+              path: {{ .Values.WebApp.HealthCheck }}
```
r4vi: need to change line 88: `readinessProble.path` too the same variable: `{{ .Values.WebApp.HealthCheck }}`

isichei: Apps seem to work with just changing livenessProbe... Is that weird / wrong?

r4vi:
No it's normal for this S3 proxy but other apps might want both probes to
point to the same healthcheck path

charts/webapp/templates/deployment.yaml
r4vi: it's normal for S3 proxy but other apps might want (and expect) both probes to
point to the same healthcheck path

PR #392, opened 2019-10-23
Rstudio: Add `spec.selector.matchLabels` to deployment

r4vi:
Unfortunately apps/v1's Deployments now needs spec.selector.matchLabels in order to work.
How did it even work before?

bug fix: https://trello.com/c/HfICQ15Y/386-add-matchselector-to-tools-charts

### DIFF
--- a/charts/rstudio/templates/deployment.yml
+++ b/charts/rstudio/templates/deployment.yml
@@ -151,3 +151,6 @@ spec:
         - name: home
           persistentVolumeClaim:
             claimName: nfs-home
+  selector:
+    matchLabels:
+      app: "{{ .Chart.Name }}"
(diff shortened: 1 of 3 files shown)


charts/rstudio/templates/deployment.yml
```diff
@@ -151,3 +151,6 @@ spec:
         - name: home
           persistentVolumeClaim:
             claimName: nfs-home
+  selector:
+    matchLabels:
+      app: "{{ .Chart.Name }}"
```
xoen: I know it's technically valid YAML and everything but any reason for hiding this at the bottom? I would have personally put that just above the `spec.template` and the pod spec labels as these two sections are interconnected.

No _need_ to change tho.

xoen:
@djsd123 and my bad for not thinking about that during code review. To be honest it's just a small sneaky change.

Thanks for fixing it @r4vi 🌟

charts/rstudio/templates/deployment.yml
r4vi: I didn't think about it, I just let my editor add it wherever it wanted to (it came up with red squiggly line saying "selector" missing so I let clippy add it wherever).
image

PR #396, opened 2019-10-28
New Env Init Platform

djsd123:
Modifying to work with serjNAS

- Modified expected mount from `/users/homes` to `/users/home`
- Modified working directory from `/tmp/home` to `/tmp`

__Deployed__

### DIFF
--- a/charts/init-platform/Chart.yaml
+++ b/charts/init-platform/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for Kubernetes
 name: init-platform
-version: 0.1.0
+version: 1.0.0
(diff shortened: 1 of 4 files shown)


charts/init-platform/Chart.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for Kubernetes
 name: init-platform
-version: 0.1.0
```
xoen: probably worth a mention in the `CHANGELOG`...

So this is necessary because SerjNAS was assuming users' homes were in `/users/home`? Would it potentially break in old clusters/SoftNAS? In that case probably better to at the very least bump the major version (or potentially make that an argument of the chart but probably overkill in our case).

djsd123: Correct. no, would technically break the in SoftNAS just create another directory.  I'll bump the major version because of this

djsd123:
Added Changelog and bumped the major version

PR #398, opened 2019-10-30
Tools play well with long domains/hosts

xoen:
Label values can't be longer than 63 chars but also can't end with
a dot (`.`) so truncate to 63 chars can lead to invalid labels.

To maintain compatibility with existing clusters, chart and unidler:
- we use full host if its lenght is max 63 chars (as before)
- we only use the part of the host before the first dot (`.`) otherwise

Updated charts:
- airflow-sqlite
- jupyter-lab
- rstudio

Charts where "host" label was not used and it is not removed:
- cpanel
- webapp


xoen:
FYI: I've the rest of the helm charts changes in my local branch almost ready but I didn't push them yet to not distract from the idea behind this change.

xoen:
Thanks gentlemen ☺️

PR #400, opened 2019-11-04
unidler v4.0.3: Uses unidler `v1.0.4` (support for `unidle-key`)

xoen:
Fix for long "host" label and added `unidleKeyLabel` value to support
`UNIDLE_KEY_LABEL` configuration (unidler will stop using `host` label
in new clusters and start using `unidle-key` label)

Part of ticket: https://trello.com/c/kVT0QFqe

### DIFF
--- a/charts/unidler/CHANGELOG.md
+++ b/charts/unidler/CHANGELOG.md
@@ -5,6 +5,26 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
+## [v4.0.3] - 2019-11-04
+### Changed
+- uses [unidler v1.0.4](https://github.com/ministryofjustice/analytics-platform-go-unidler/releases/tag/v1.0.4)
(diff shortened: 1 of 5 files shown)


charts/unidler/CHANGELOG.md
```diff
@@ -5,6 +5,26 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
+## [v4.0.3] - 2019-11-04
+### Changed
+- uses [unidler v1.0.4](https://github.com/ministryofjustice/analytics-platform-go-unidler/releases/tag/v1.0.4)
+- this version calculate "unidle key" slightly differently
+  to avoid problems with new long domain: it only uses
+  the part of the host before the first dot (".") when
+  the host is longer than 63 character
```
djsd123: Is that right?  `it only uses the part of the host before the first dot (".") when  the host is longer than 63 character`

I thought it used the prefix before the period (`.`) regardless.

xoen:
Thanks Mikael 🥇

PR #401, opened 2019-11-05
SerjNAS NFS Path

djsd123:
Breaking Change
---------------
Changed `PersistentVolume` NFS path from `/users/homes/{{
.Values.Username }}` to `"/users/home/{{ .Values.Username }}"`

Forgot to do this 😞


xoen:
@sh31 does this look good to you?

djsd123:
@sh31 has modified the NFS export from `/users/home` to `/users/homes` 🤷‍♂.

PR #403, opened 2019-11-05
Auth0 User Exports

djsd123:
[Trello](https://trello.com/c/0b73fK0F)

Helm Chart for Auth0 User exports utility

Configured to export data to `S3`

### DIFF
--- a/charts/auth0-user-exports/Chart.yaml
+++ b/charts/auth0-user-exports/Chart.yaml
@@ -1,0 +1,7 @@
+apiVersion: v1
+appVersion: "0.1.0"
+description: Query Auth0's API and export data in csv format to either local file system or an S3 bucket
+name: auth0-user-exports
+version: 0.1.0
+sources:
+  - https://github.com/ministryofjustice/analytical-platform-auth0-user-exports
--- a/charts/auth0-user-exports/README.md
+++ b/charts/auth0-user-exports/README.md
@@ -1,0 +1,27 @@
+# Auth0 User Exports
+
+##### CronJob that runs [auth0-user-exports](https://github.com/ministryofjustice/analytical-platform-auth0-user-exports)
+
+### Install/Upgrade
+
+```bash
+helm upgrade auth0-user-exports mojanalytics/auth0-user-exports --namespace default --values ../analytics-platform-config/chart-env-config/$ENV/auth0-user-exports --install
+```
+
+### Remove
+
+```bash
+helm del --purge auth0-user-exports
+```
+
+### Configuration
+
+| Parameter  | Description      | Default |
+| ---------- | ---------------  | ------- |
+| `schedule` | When the job should execute (In **Cron** format) | *Monday to Friday at 23:00* |
+| `env.auth0_domain` | Auth0 management API endpoint | https://alpha-analytics-moj.eu.auth0.com |
+| `env.auth0_client_id` | The Client ID of the auth0 application used for this app | "" |
+| `env.auth0_client_secret` | The Client Secret of the auth0 application used for this app | "" |
+| `env.env` | **OPTIONAL** Environment in which the app is running. Set to `aws` to write to s3 | `aws` |
+| `env.role_arn` | The `IAM` role to assume when writing to `s3` | `arn:aws:iam:::role/auth0-user-data-write` |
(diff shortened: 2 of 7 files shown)


charts/auth0-user-exports/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+appVersion: "0.1.0"
+description: Query Auth0's API and export data in csv format to either local file system or an S3 bucket
+name: auth0-user-exports
+version: 0.1.0
+sources:
+  - https://github.com/ministryofjustice/analytical-platform-auth0-user-exports
```
xoen: nice, didn't know about `sources`, very useful, I'll add to the other helm charts from now on 🏆

charts/auth0-user-exports/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+```bash
+helm del --purge auth0-user-exports
+```
+
+### Configuration
+
+| Parameter  | Description      | Default |
+| ---------- | ---------------  | ------- |
+| `schedule` | When the job should execute (In **Cron** format) | *Monday to Friday at 23:00* |
```
xoen: I suppose Sam doesn't care about gaps in the data for weekends 🤷‍♂

charts/auth0-user-exports/templates/NOTES.txt
```diff
@@ -0,0 +1,21 @@
+1. Get the application URL by running these commands:
```
xoen: What is this? Isn't a pod for the `CronJob` only there briefly when it's run at the given schedule?

charts/auth0-user-exports/templates/tests/test-connection.yaml
```diff
@@ -0,0 +1,18 @@
...
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+  annotations:
+    "helm.sh/hook": test-success
+spec:
+  containers:
+    - name: wget
+      image: busybox
+      command: ['wget']
+      args:  ['{{ include "auth0-user-exports-chart.fullname" . }}:{{ .Values.service.port }}']
```
xoen: how is this working? I'm probably missing something here but what's wget-tting here if the pod is only there when the `Job` is executed?

xoen:
Helm chart looks good (but haven't looked at the code the `Job` is running yet) but I don't get what's the notes and the helm test are about.

Looking at this made me wonder if the job should instead be run by Airflow instead of kubernetes. Reasons why:
1. it gathers data and put into an S3 bucket for later analysis, it's basically doing the Extraction bit in a ETL process
2. it would give Sam/the Data Engineers team visibility over when this run or failed, etc...
3. one less thing the Real Ops have to worry about (whether it run or not, etc...)

Most of our Airflow DAGs are very similar (conceptually) to this anyway, e.g. they execute a docker container on a certain schedule. Maybe show this to @s-block and explain what it's doing and see if it would make sense to run the user export in Airflow.

charts/auth0-user-exports/templates/NOTES.txt
djsd123: Sorry, that's boilerplate.  Will remove

charts/auth0-user-exports/templates/tests/test-connection.yaml
djsd123: Same

charts/auth0-user-exports/README.md
djsd123: I figured it would be unlikely for new users to join the platform over the weekend 🤷‍♂

charts/auth0-user-exports/Chart.yaml
djsd123: https://helm.sh/docs/developing_charts/#the-chart-yaml-file

djsd123:
Closing this in favour of airflow DAG.  Cheers @xoen

charts/auth0-user-exports/Chart.yaml
xoen: Yeah saw that after I saw this PR, such a Pro 😎

PR #407, opened 2019-11-13
idler 2.1.0: Run idler every Wednesday by default

xoen:
This is to reduce users' waiting time in the morning.

I've updated the default `schedule` as we currently we don't have config
files for this helm chart and it makes sense to keep installation of
this chart simple rather then require a config file.

Ticket: https://trello.com/c/6yyF8WFV


xoen:
FYI: Manually edited `Job`'s `schedule` in `alpha` so that we can monitor this for a week or so and then decide if we're still happy with idling only on Wednesday nights.

For the record: In general we should avoid manually edit k8s resources but in this case makes sense to test it first and also it's a very low impact/risk change.

xoen:
FYI: changed from "every Wednesday night" to "every **Tuesday** night" as discussed, re-deployed both in `dev` and `alpha`. Merging.

PR #411, opened 2019-11-25
Update airflow image

s-block:


xoen:
@s-block looks good to me but I don't see the image tag/release https://github.com/ministryofjustice/analytics-platform-airflow-docker-image/releases

Was this pushed manually to quay? We should avoid that (assuming that's what happened), could you create a release in the GH repository as well please? Thanks ☺️

s-block:
https://github.com/ministryofjustice/analytics-platform-airflow-docker-image/releases/tag/1.10.6-1

xoen:
Yep ignore me, I saw the releases but missed that specific one because it was AFTER the `1.10.6` release because of weird ordering ☺️

PR #413, opened 2019-12-16
Tools fixed memory

sh31:
* Fixed memory and CPU for R Studio and Jupyter Lab . This sets [K8S QOS](https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/) to guaranteed.

### DIFF
--- a/charts/jupyter-lab/Chart.yaml
+++ b/charts/jupyter-lab/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
 description: Jupyter Lab with Auth0 authentication proxy
 name: jupyter-lab
-version: 0.4.6
+version: 0.4.7
 appVersion: v0.6.7
--- a/charts/jupyter-lab/Chart.yaml
+++ b/charts/jupyter-lab/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
 description: Jupyter Lab with Auth0 authentication proxy
 name: jupyter-lab
-version: 0.4.7
+version: 0.5.0
 appVersion: v0.6.7
--- a/charts/rstudio/CHANGELOG.md
+++ b/charts/rstudio/CHANGELOG.md
@@ -5,6 +5,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+## [2.1.6] - 2020-01-21
+### Changed
+Give containers fixed cpu and memory to guarantee QoS
+
+
 ## [2.1.5] - 2019-10-30
 ### Changed
 Diffent approach to long `host` labels.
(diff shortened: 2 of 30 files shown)


xoen:
I'm having a look at that k8s documentation link you provided, doesn't that mean you also need to adjust the proxies resources as well?

> **Every Container** in the Pod must have a memory/CPU limit and a memory/CPU request, and they must be the same

(Emphasis mine)

charts/jupyter-lab/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
 description: Jupyter Lab with Auth0 authentication proxy
 name: jupyter-lab
-version: 0.4.6
+version: 0.4.7
```
xoen: Don't know if it's worth bumping the MINOR instead of the PATCH in this case. I suppose depends on whether we consider switching to QoS `Guaranteed ` a feature or a bug-fix:

From https://semver.org/:

> Given a version number MAJOR.MINOR.PATCH, increment the:
> MAJOR version when you make incompatible API changes,
> MINOR version when you add functionality in a backwards compatible manner, and
> PATCH version when you make backwards compatible bug fixes.

charts/rstudio/CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+## [2.1.6] - 2020-01-21
```
xoen: Again (see [other comment](https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/413/files#r368958461)) this may deserves a bump to `2.2.0`.

xoen:
Seal of Approval

charts/rstudio/values.yaml
```diff
@@ -39,8 +39,8 @@ rstudio:
     init: true
   resources:
     limits:
-      cpu: 1.5
-      memory: 20Gi
+      cpu: 200m
+      memory: 12Gi
```
xoen: As discussed, most users are using ~10GB of memory AFAICT from Grafana - unless this is unusual set the limit to 12GB may be OK. Increasing this may require even more k8s nodes and many users don't use this much memory anyway so 12GB may be a good compromise.

charts/webapp/templates/fluent-bit-configmap.yml
```diff
@@ -6,6 +6,7 @@ metadata:
     app: {{ template "fullname" . }}
     heritage: "{{ .Release.Service }}"
     release: "{{ .Release.Name }}"
+    host: {{ include "hostname" . | trunc 63 }}
```
xoen: This is weird. I don't know what went wrong with the `git rebase` this change shouldn't be here as these are already in `master` - unless GitHub diff got confused.

PR #414, opened 2020-01-06
webapp 2.3.7: Revert "webapp 2.3.5: removed unused `host` labels"

xoen:
This reverts commit 04cd02233449af29700e126090732fe186f75f38 (so basically I'm adding back `host` label for webapps)

The `host` label is not only used by the [*restarter* service](https://github.com/ministryofjustice/analytics-platform-restarter) but
it's also causing some webapp deployments to fail because
of a new constraint on labels in `apps/v1` which means
they can't change

SEE: [kubernetes issue on GitHub](https://github.com/kubernetes/kubernetes/issues/26202)

Ticket: https://trello.com/c/drbbqgjC


xoen:
FYI: restarter only needs the `host` label to be in the `Deployment` but I tried to not be too clever and just reverted it with the only tweak to changelog and helm chart version.

PR #416, opened 2020-01-22
Remove heapster

davidread:
We use metrics-server since Sept 2018
Metrics-server chart install is documented here:
https://github.com/ministryofjustice/analytics-platform-config/blob/master/chart-env-config/alpha/metrics-server.yaml


davidread:
👍

PR #419, opened 2020-02-28
Call nbstripout to remove outputs from .ipynb files

dhirajnarwanimoj:
For each .ipynb file commited to a git repo. On pre-commit

Call nbstripout (https://github.com/kynan/nbstripout) 
This will remove the outputs from the file 
And then add the file to the repo

This is as part of story (https://trello.com/c/P7BRQJWC/333-strip-out-ipython-notebook-output-using-a-git-hook-for-all-users)

### DIFF
--- a/charts/config-user/files/pre-commit
+++ b/charts/config-user/files/pre-commit
@@ -39,6 +39,7 @@ NBS=`git diff-index -z --cached $against --name-only | grep '.ipynb$' | uniq`
 for NB in $NBS ; do
     echo "Removing outputs from $NB"
     nbstripout "$NB"
+    echo "nbstripout has run"
     git add "$NB"
 done
 )
--- a/charts/config-user/files/pre-commit
+++ b/charts/config-user/files/pre-commit
@@ -34,12 +34,13 @@ fi
 (
 IFS='
 '
-NBS=`git diff-index -z --cached $against --name-only | grep '.ipynb$' | uniq`
+NBS=`git diff --cached $against --name-only | grep '.ipynb$' | sort | uniq`
 
 for NB in $NBS ; do
     echo "Removing outputs from $NB"
     nbstripout "$NB"
     echo "nbstripout has run"
+
     git add "$NB"
 done
 )
--- a/charts/config-user/files/pre-commit
+++ b/charts/config-user/files/pre-commit
@@ -20,3 +20,27 @@ if [ "$HAS_ERROR" != "" ]; then
     echo "Can't commit due to oversize files, please unstage before committing." >&2
     exit 1
 fi
+
+
+# add nbstripout to strip output from IPython notebooks 
+if git rev-parse --verify HEAD >/dev/null 2>&1; then
+   against=HEAD
+else
+   # Initial commit: diff against an empty tree object
+   against=4b825dc642cb6eb9a060e54bf8d69288fbee4904
+fi
+
+# Find notebooks to be committed
+(
+IFS='
+'
+NBS=`git diff-index -z --cached $against --name-only | grep '.ipynb$' | uniq`
+
+for NB in $NBS ; do
+    echo "Removing outputs from $NB"
+    nbstripout "$NB"
+    git add "$NB"
+done
+)
+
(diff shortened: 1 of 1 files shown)


charts/config-user/files/pre-commit
```diff
@@ -20,3 +20,29 @@ if [ "$HAS_ERROR" != "" ]; then
...
+if git rev-parse --verify HEAD >/dev/null 2>&1; then
+   against=HEAD
+else
+   # Initial commit: diff against an empty tree object
+   against=4b825dc642cb6eb9a060e54bf8d69288fbee4904
+fi
+
+# Find notebooks to be committed
+(
+IFS='
```
xoen: what's `IFS`? My bash is not great but it doesn't seem to be used (assuming is a variable, which may well not be)?

charts/config-user/files/pre-commit
```diff
@@ -20,3 +20,29 @@ if [ "$HAS_ERROR" != "" ]; then
...
     exit 1
 fi
+
+
+# add nbstripout to strip output from IPython notebooks 
+if git rev-parse --verify HEAD >/dev/null 2>&1; then
+   against=HEAD
+else
+   # Initial commit: diff against an empty tree object
+   against=4b825dc642cb6eb9a060e54bf8d69288fbee4904
```
xoen: Magic. Never heard of this one.

charts/config-user/files/pre-commit
```diff
@@ -20,3 +20,29 @@ if [ "$HAS_ERROR" != "" ]; then
...
+IFS='
+'
+NBS=`git diff --cached $against --name-only | grep '.ipynb$' | sort | uniq`
+
+for NB in $NBS ; do
+    echo "Removing outputs from $NB"
+    nbstripout "$NB"
+    echo "nbstripout has run"
+
+    git add "$NB"
```
xoen: never had to write a Git pre-commit script, how does it work? If you do `git add` it will add the object to the commit the rest of the files will not end up in the commit?

Sorry for the stupid question.

charts/config-user/files/pre-commit
```diff
@@ -20,3 +20,29 @@ if [ "$HAS_ERROR" != "" ]; then
...
+done
+)
+
+exec git diff-index --check --cached $against --
```
xoen: Never used `git diff-index`.

From quick look at the manpage is basically a diff but you pass `--check`:

> Warn if changes introduce conflict markers or whitespace errors.
> [...]
> Exits with non-zero status if problems are found.

So is this to fail the commit if somehow this introduces a conflict?
It's not clear to me, maybe worth adding a small note/comment for whoever is gonna look at it in the future :)

charts/config-user/files/pre-commit
dhirajnarwanimoj: Yes so you create a global pre-commit hook. 
This hook will be added to the pre-commit inside the user's machine. 

And then whenever the user commits on his local machine 
it will check the hook. 
in this case in checks the file extension .ipyb and then it will call nbstripout on that file and then add it to git

charts/config-user/files/pre-commit
dhirajnarwanimoj: Basically IFS separates the word in 1 line so when you iterate the for loops reads it as single objects

charts/config-user/files/pre-commit
xoen: Right. I wasn't sure about the `git add` part.
So basically:
- the `*.ipynb` was staged
- nbstripout updates the file
- and `git add` re-stage the modified version.

Cool, makes sense 👍

charts/config-user/files/pre-commit
dhirajnarwanimoj: i don't think we need this actually 
i'll remove it

charts/config-user/files/pre-commit
xoen: Oh I never heard of it, more Bash-fu 🤣

davidread: Amazing, I love learning about this sort of stuff!

PR #422, opened 2020-03-11
set git-template path to ~/.git-templates

dhirajnarwanimoj:
Set global git template path to ~/.git-templates and not USER_HOME (dhirajnarwanimoj).
The reason is that jupyter runs another user (jovyan) and we need to change the ~/.git-templates of that user


charts/config-user/files/git-config.sh
```diff
@@ -56,5 +56,5 @@ git config -f $GIT_CONFIG user.email $EMAIL
 git config -f $GIT_CONFIG user.name "${FULLNAME}"
 git config -f $GIT_CONFIG core.excludesfile $GIT_IGNORE
 git config -f $GIT_CONFIG init.templatedir $GIT_TEMPLATES
```
davidread: Is it a problem also for init.templatedir being set to the wrong directory when accessed in jupyter?

dhirajnarwanimoj: well i don't think this is an issue because it's working

PR #424, opened 2020-03-12
revert back to jovyan user

dhirajnarwanimoj:
Revert back the git user to jovyan

### DIFF
--- a/charts/config-user/Chart.yaml
+++ b/charts/config-user/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: Configure the user (e.g. git setup, etc...)
 name: config-user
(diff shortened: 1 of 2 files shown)


davidread:
I appreciate you're iterating, but for someone to review a PR it would be helpful if you can explain the background, for now and the next person looking at this line.

* Why are you having to set the user in the first place?
* Why didn't 1001 work?
* How are you going to address people committing from rstudio?

One thought - have you considered setting this option in git's [System-wide configuration file](https://git-scm.com/docs/git-config#FILES) in the R Studio and Jupyter dockerfile, so it is set on pod start-up, rather than editing every user's home directory?

PR #425, opened 2020-03-13
Set git global

dhirajnarwanimoj:
The issue is that .git-templates was set to root and the hooks folder wasn't getting copied.

We need to chown to 1001:staff to .git-templates

### DIFF
--- a/charts/config-user/Chart.yaml
+++ b/charts/config-user/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: Configure the user (e.g. git setup, etc...)
 name: config-user
-version: 0.2.5
+version: 0.2.6
--- a/charts/config-user/files/git-config.sh
+++ b/charts/config-user/files/git-config.sh
@@ -58,4 +58,4 @@ git config -f $GIT_CONFIG core.excludesfile $GIT_IGNORE
 git config -f $GIT_CONFIG init.templatedir $GIT_TEMPLATES
 
 chown 1001:staff $GIT_CONFIG
-chmod 1001:staff $GIT_TEMPLATES
+chown 1001:staff $GIT_TEMPLATES
--- a/charts/config-user/files/git-config.sh
+++ b/charts/config-user/files/git-config.sh
@@ -56,8 +56,6 @@ git config -f $GIT_CONFIG user.email $EMAIL
 git config -f $GIT_CONFIG user.name "${FULLNAME}"
 git config -f $GIT_CONFIG core.excludesfile $GIT_IGNORE
 git config -f $GIT_CONFIG init.templatedir $GIT_TEMPLATES
-chown 1001:staff $GIT_CONFIG
 
-# set global git template config for jupyter user and git user (rstudio)
-git config --global init.templatedir '/home/jovyan/.git-templates'
-git config --global init.templatedir $GIT_TEMPLATES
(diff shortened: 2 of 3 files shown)


charts/config-user/files/git-config.sh
```diff
@@ -56,16 +56,6 @@ git config -f $GIT_CONFIG user.email $EMAIL
...
-  USER=$USERNAME
-fi
 
-# set global templates either with the jupyter or git user
-if [ $USER ]; then
-  sudo -u $USER git config --global init.templatedir '~/.git-templates'
-fi
\ No newline at end of file
+chown 1001:staff $GIT_CONFIG
+chmod 1001:staff $GIT_TEMPLATES
```
xoen: can `chmod` be used like this? I thought he needed the desired mode or permission change, e.g. `chmod u+rw ...`.

Anyway, is changing the permissions necessary? can't you tweak the the command above to be 
recursive and update the owner on the user home?

(something like this)

```
chown -R 1001:staff $USER_HOME
```

@sh31 could you have a look at this when you get the time please?

dhirajnarwanimoj: yeah so we have the group set up with the appropriate permissions, so all you need to use that group and user.

You can't set USER_HOME because I think some need to have root access

dhirajnarwanimoj: oh you are right! Woops it needs to be chown

charts/config-user/files/git-config.sh
```diff
@@ -56,16 +56,6 @@ git config -f $GIT_CONFIG user.email $EMAIL
...
-  USER=$USERNAME
-fi
 
-# set global templates either with the jupyter or git user
-if [ $USER ]; then
-  sudo -u $USER git config --global init.templatedir '~/.git-templates'
-fi
\ No newline at end of file
+chown 1001:staff $GIT_CONFIG
+chown 1001:staff $GIT_TEMPLATES
```
xoen: don't you still need this command to be recursive (adding `-R`) when it's a directory? (I may be wrong)

charts/config-user/Chart.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: Configure the user (e.g. git setup, etc...)
 name: config-user
-version: 0.2.5
+version: 0.2.6
```
xoen: I would also add a note to the changelog file of the helm chart as well, to keep track of these changes.

charts/config-user/files/git-config.sh
dhirajnarwanimoj: no because 1001 already has the R permission

charts/config-user/Chart.yaml
dhirajnarwanimoj: there's no CHANGELOG, let me add one

charts/config-user/files/git-config.sh
sh31: Echoing what @xoen said. I think you may be seeing that 1001 already has the permission as you are manually creating the files in the directory as that user for testing?

Have you tested this by applying the Helm chart rather than creating the files yourself?

dhirajnarwanimoj: No I added the chmod manually and it worked. I need to push this to master to test the helm chart

PR #426, opened 2020-03-17
change  to ~/.git-templates

dhirajnarwanimoj:
Apparently the current $GIT_TEMPLATES points to /home/user/.git-templates.
This won't work on jupyter as it uses /home/jovyan/.git-templates

We need to set the init.templatedir to ~/.git-templates and chown that to 1001:staff

Also we need to add the full path to nbstripout in the pre-commit hook file as in r-studio it doesn't find the package.


dhirajnarwanimoj:
> I don't understand this.
> 
> The `config-user` job mounts the user home in `$USER_HOME` which is eventually mounted in `/home/jovian` (for JupyterLab) so these two are the same.
> 
> @dhirajnarwanimoj
> 
> 1. could you explain better what was the error you were seeing that prompted you to change this?
> 2. are you sure that the `chown` of the `$GIT_TEMPLATES` directory doesn't need the `-R` (recursive) flag?
> 3. isn't `~` evaluating to the `$USER_HOME`/`/home/${USERNAME}` anyway? what difference is going to make using `~/.git-templates`?
> 
> Could you be **specific** about what you have tried/done and what was the output/eventual error that prompted you to make these changes?

When this ran - git config init.templatedir $USER_HOME/.git-templates. I checked on the terminal and it was set to my username (dhirajnarwanimoj) not jovyan and that's why this fix didn't work. 

So if we set it to ~/.git-templates - this automatically will set it to local user of that terminal - in this case jovyan

xoen:
> When this ran - git config init.templatedir $USER_HOME/.git-templates. I checked on the terminal and it was set to my username (dhirajnarwanimoj) not jovyan and that's why this fix didn't work.

What was the context? Where did you run `git config init.templatedir $USER_HOME/.git-templates`? Bear in mind when these commands run in the `config-user` job the environment, environment variables, etc are different than your own JupyterLab environment.

davidread:
Yes we could do with an explanation. I'll provide my understanding and perhaps @dhirajnarwanimoj can correct any wrong bits.

When the helm chart runs `git config init.templatedir $USER_HOME/.git-templates` it simply writes that into ~/.gitconfig. I assume it would resolve the environment variable at that point, so it saves is saved as `init.templatedir = /home/davidread/.git-templates`.

When the user goes to an R Studio or Jupyter terminal and runs `git init` it tries to copy the templates from that configured dir into the the local .git directory. This succeeds on R Studio, but of course on Jupyter it can't find `/home/davidread` at all - it needs to get it from `/home/jovyan/.git-templates`, because Jupyter mounts the home dir in that other place. So the idea tried here is to refer to the home directory as `~`, so that when it is resolved at the time of running `git init` it works in both environments.

xoen:
@davidread thanks for the explanation 💯 - so the git configuration will have the `~` which will be expanded at the time of running `git init` which means it will work regardless of where the home is actually mounted. Now it makes more sense 👌

PR #427, opened 2020-03-19
Tag r-studio-server with 1.2.1335-r3.5.1-python3.7.1-conda-4

dhirajnarwanimoj:
We need to set the r-studio-tag to a new tag '1.2.1335-r3.5.1-python3.7.1-conda-4'. 

This has been tested and released already on quay

### DIFF
--- a/charts/rstudio/CHANGELOG.md
+++ b/charts/rstudio/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [2.2.2] - 2020-03-19
+### Changed
+Changed tag to latest
+Bumped version
+
+See:
+- https://github.com/ministryofjustice/analytics-platform-auth-proxy/releases/tag/v5.2.4
+- https://github.com/ministryofjustice/analytics-platform-auth-proxy/releases/tag/v5.2.3
 
 ## [2.2.1] - 2020-01-31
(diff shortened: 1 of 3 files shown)


dhirajnarwanimoj:
> Sorry Dhiraj but this is not a best practice, **we should not have any docker image tag pointing to `latest`** because this is really unpredictable ❌
> 
> In addition to that, when using `latest` it means that k8s will not re-create a pod or pull the new version of an image because it already has an image with that tag (which is obviously different tho).
> Even change k8s `pullPolicy` to always is bad because it would be wasteful and trigger the re-pulling of a docker image all the time which is even worse in the case of large docker images like the tools ones.
> 
> CC: Maybe @sh31 or @davidread have something to add to this.

So should I tag with the last feature i've applied. In this case add-nbstripout-pkg

xoen:
@dhirajnarwanimoj yeah I think so, if no R/RStudio version or anything changed you may just bump the last bit to `1.2.1335-r3.5.1-python3.7.1-conda-4` (current version is `-3`). 
Create a release in the GH repository for RStudio called like that and check that quay.io built/created that tag [in the `rstudio` repository](https://quay.io/repository/mojanalytics/rstudio?tab=tags) ☺️

dhirajnarwanimoj:
> @dhirajnarwanimoj yeah I think so, if no R/RStudio version or anything changed you may just bump the last bit to `1.2.1335-r3.5.1-python3.7.1-conda-4` (current version is `-3`).
> Create a release in the GH repository for RStudio called like that and check that quay.io built/created that tag [in the `rstudio` repository](https://quay.io/repository/mojanalytics/rstudio?tab=tags) ☺️

So how does quay pick up the release name? Is it just the branch name?

PR #432, opened 2020-03-30
Add path to pre-commit hook

dhirajnarwanimoj:
Add Path to both jupyter and r-studio nbstripout PATHS to git pre-commit hook rather than individual docker image.

In dev nbstripout gets installed in ~/.local/bin and in alpha in /opt/conda/bin.

### DIFF
--- a/charts/config-user/CHANGELOG.md
+++ b/charts/config-user/CHANGELOG.md
@@ -1,3 +1,7 @@
+## [0.2.10] - 2020-03-31
+### Changed
+- Export PATH for rstudio and jupyter for dev and alpha environments 
+
 ## [0.2.9] - 2020-03-23
 ### Changed
 - Add -R to $GIT_TEMPLATES to set hooks as executables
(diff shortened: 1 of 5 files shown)


dhirajnarwanimoj:
> Also might be worth putting a comment about why you are changing the image tags for Jupyter and Rstudio?

This is normal whenever there is a new release, we need to bump the version up

davidread:
The reason for the chart to use these newer releases of rstudio and jupyter is because they have nbstripout installed. I agree that this should be mentioned in this PR, and ideally the CHANGELOG for this chart (create one, like other charts have?).

Additionally, please put in the release notes for jupyter https://github.com/ministryofjustice/analytics-platform-jupyter-notebook/releases what each one changes.

dhirajnarwanimoj:
> The reason for the chart to use these newer releases of rstudio and jupyter is because they have nbstripout installed. I agree that this should be mentioned in this PR, and ideally the CHANGELOG for this chart (create one, like other charts have?).
> 
> Additionally, please put in the release notes for jupyter https://github.com/ministryofjustice/analytics-platform-jupyter-notebook/releases what each one changes.

Done

PR #435, opened 2020-04-24
add spark and datascience notebook versions

dhirajnarwanimoj:
Add spark and datascience notebook versions in the comments.


andylightfoot:
No longer required

PR #439, opened 2020-05-13
Improve RStudio upgrade commands

davidread:
* I had problems doing upgrades that I'm keen for others to avoid using delete --purge.
* Seems clearer to mention the --version param than write the whole command again
* Chart resource names have changed from rstudio-USERNAME to USERNAME-rstudio

### DIFF
--- a/charts/rstudio/README.md
+++ b/charts/rstudio/README.md
@@ -1,27 +1,28 @@
 # RStudio Helm Chart
 
-
 ## Install/upgrade the Chart
 
 To install/upgrade an rstudio instance for the user specified in the username variable (Github username):
 
 ```bash
-$ helm upgrade --dry-run --debug --install USERNAME-rstudio --namespace user-USERNAME --set username=USERNAME --set aws.iamRole=ENV_user_USERNAME charts/rstudio -f chart-env-config/ENV/rstudio.yml
+helm upgrade --dry-run --debug --install USERNAME-rstudio --namespace user-USERNAME --set username=USERNAME --set aws.iamRole=ENV_user_USERNAME charts/rstudio -f chart-env-config/ENV/rstudio.yml
 ```
 
-or if you want to upgrade/downgrade the helm chart to a version from the helm repository you can do the following:
+The instance will be available at <https://USERNAME-rstudio.tools.ENV.mojanalytics.xyz>.
+
+### Notes
+
+Change the environment config file to deploy in a different environment (the URL will change accordingly)
+
+Remove the `--dry-run` option to install/upgrade for real.
+
+If you want a particular version, add parameter: ``--version CHART_VERSION`` e.g. ``--version 2.2.5``.
+
+NB helm is [not good at doing upgrades](https://github.com/ministryofjustice/analytics-platform/wiki/R-Studio-start-up-failure#fix-with-helm) - if you run into trouble, it's usually easiest to delete+purge the chart and install again:
 
 ```bash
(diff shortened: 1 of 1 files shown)


charts/rstudio/README.md
```diff
@@ -1,27 +1,28 @@
 # RStudio Helm Chart
 
-
 ## Install/upgrade the Chart
 
 To install/upgrade an rstudio instance for the user specified in the username variable (Github username):
 
 ```bash
-$ helm upgrade --dry-run --debug --install USERNAME-rstudio --namespace user-USERNAME --set username=USERNAME --set aws.iamRole=ENV_user_USERNAME charts/rstudio -f chart-env-config/ENV/rstudio.yml
+helm upgrade --dry-run --debug --install USERNAME-rstudio --namespace user-USERNAME --set username=USERNAME --set aws.iamRole=ENV_user_USERNAME charts/rstudio -f chart-env-config/ENV/rstudio.yml
```
xoen, line 8: The `$` prompt is an old convention that this is a shell command you can run as non-root (as opposite to `#`), it give the reader of this an hint this is a shell command and not, let's say python or something else (I know it sounds 100% obvious but it always help to be clear)

We've been using this in other READMEs and wiki parts AFAICT.

charts/rstudio/README.md
```diff
@@ -1,27 +1,28 @@
...
 ```
 
-or if you want to upgrade/downgrade the helm chart to a version from the helm repository you can do the following:
+The instance will be available in <https://USERNAME-rstudio.tools.ENV.mojanalytics.xyz>.
```
xoen, line 11: Not a native English speaker but isn't "at" instead of "in" the correct preposition in this case? 🤷‍♂️

charts/rstudio/README.md
```diff
@@ -1,27 +1,28 @@
...
 
-```bash
-$ helm upgrade --dry-run  --debug --install USERNAME-rstudio mojanalytics/rstudio --namespace user-USERNAME --set username=USERNAME --set aws.iamRole=ENV_user_USERNAME --version CHART_VERSION -f chart-env-config/ENV/rstudio.yml
```
xoen, line 15: I notice that you got rid of the `helm upgrade` command.

We tend to at least have the `helm upgrade --install` command in these READMEs as they're the most versitile, e.g. you can use the same command to install and re-install/upgrade instead of having to tweak the `helm install` command.

Probably better to get rid of the `helm install` command above and replace with the correct `helm upgrade --install` command.

charts/rstudio/README.md
```diff
@@ -1,27 +1,28 @@
...
 
-**NOTE**: Change the environment config file to deploy in a different environment
-          (the URL will change accordingly)
+If you want a particular version, add parameter: ``--version CHART_VERSION`` e.g. ``--version 2.2.5``.
 
-**NOTE**: Remove the `--dry-run` option to install/upgrade for real.
+NB helm is [not good at doing upgrades](https://github.com/ministryofjustice/analytics-platform/wiki/R-Studio-start-up-failure#fix-with-helm) - if you run into trouble, it's usually easiest to delete+purge the chart and install again:
 
+```bash
+helm delete --purge rstudio-USERNAME
```
xoen, line 24: 👌 

I'd add the `$ ` prefix tho.

charts/rstudio/README.md
davidread: I think showing the prompt is definitely helpful when you're separating a command from output. Otherwise I prefer to avoid it because it looks a bit dated and there is the usability annoyance when you double click to copy when you're pasting to the command-line.

charts/rstudio/README.md
davidread: It is already how you want it, I believe?

PR #440, opened 2020-05-15
Update version of auth-proxy to 5.2.5

ntoll:
Does what it says on the tin.


ntoll:
@andylightfoot doh..! Thanks. Done.

PR #441, opened 2020-06-15
Use v5.2.6 of auth-proxy in helm charts

xoen:
This is a maintenance release of the `auth-proxy` which upgrades
some of its dependencies.
Few of these new dependencies also fix some potential security
vulnerabilities.

This commit updates the following helm charts to use this new version,
all the helm charts were using a recent version of the `auth-proxy`
of the `v5.2.x` series so this change should not - in theory - break any
of the existing helm charts in theory:

- airflow-sqlite
- jupyter-lab
- kibana-auth-proxy
- kubernetes-dashboard
- prometheus-resources
- rstudio
- webapp

See `auth-proxy` Release Notes for more info: https://github.com/ministryofjustice/analytics-platform-auth-proxy/releases/tag/v5.2.6

Part of ticket: https://trello.com/c/wTQiRlcJ

### DIFF
--- a/charts/airflow-sqlite/CHANGELOG.md
+++ b/charts/airflow-sqlite/CHANGELOG.md
@@ -5,6 +5,17 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+## [0.3.2] - 2020-06-15
+### Changed
+Bumped auth-proxy from `v5.2.5` to `v5.2.6`.
+
+This is a maintenance release which upgrades some of its dependencies.
+Few of these new dependencies also fix some potential security
+vulnerabilities.
+
+See Release Notes for more: https://github.com/ministryofjustice/analytics-platform-auth-proxy/releases/tag/v5.2.6
+
+
 ## [0.3.1] - 2020-01-31
 ### Changed
 Bumped auth-proxy from `v5.2.1` to `v5.2.4`.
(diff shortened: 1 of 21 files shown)


ntoll:
@xoen Why shouldn't this be merged yet..?

xoen:
@ntoll - I've tested the proxy with a couple of simple webapps but I want to do some more manual testing for the RStudio/JupyterLab ones before I merge this by making these new versions of the charts available to all users. 

I don't expect anything to go wrong really but it's just caution just in case, as once this is merged new versions of the charts are created, etc...

xoen:
@andylightfoot I think I should be able to install the newer versions of these three helm charts:

- kibana-auth-proxy
- kubernetes-dashboard
- prometheus-resources

but I may need some potential hand holding as I'm not that familiar with these and I feel like blindly upgrading those charts without knowing much about those could lead to some problems - nice thing of helm is that we should 9/10 be able to rollback to working revision if anything goes wrong but still...

I may bug you 🐞tomorrow.

xoen:
FYI: I've tested this new version of the proxy for my `dev` instances of `rstudio`, `jupyter-lab` and `airflow-sqlite`. Bear in mind my testing was pretty shallow - logged in and clicked around - but it's fair to say the proxy seems to work as before for the tools.

xoen:
FYI: Upgraded/tested the [non-tools helm releases](https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/441#issuecomment-644211928) in `dev` and they appear to be working as before with new proxy version (or at least based from quick inspection)

xoen:
FYI: I'm merging this so that users can start using new version of tools helm charts.
I'll deploy the system-charts in `alpha` later (as I already said I've done that in `dev`).

PR #442, opened 2020-07-02
Trello-564-alert-settings

tomskelley-gjs:
Add kubernetesSystem monitoring checks, but with custom timing for KubeNodeNotReady which is hardcoded in the public chart

### DIFF
--- a/charts/prometheus-resources/values.yaml
+++ b/charts/prometheus-resources/values.yaml
@@ -72,3 +72,103 @@ prometheusRules:
         for: 1m
         label:
           severity: warning
+
+# custom kubernetes rules
+    - name: kubernetes-system
+      rules:
+      - alert: KubeNodeNotReady
+        annotations:
+          message: '{{`{{ $labels.node }}`}} has been unready for more than 5 minutes.'
+          runbook_url: https://github.com/kubernetes-monitoring/kubernetes-mixin/tree/master/runbook.md#alert-name-kubenodenotready
+        expr: kube_node_status_condition{job="kube-state-metrics",condition="Ready",status="true"} == 0
+        for: 5m
+        labels:
+          severity: critical
+      - alert: KubeVersionMismatch
+        annotations:
+          message: There are {{`{{ $value }}`}} different semantic versions of Kubernetes components running.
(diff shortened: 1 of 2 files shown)


charts/prometheus-resources/values.yaml
```diff
@@ -72,3 +72,103 @@ prometheusRules:
...
+        annotations:
+          message: Kubernetes API server client '{{`{{ $labels.job }}`}}/{{`{{ $labels.instance }}`}}' is experiencing {{`{{ printf "%0.0f" $value }}`}} errors / second.
+          runbook_url: https://github.com/kubernetes-monitoring/kubernetes-mixin/tree/master/runbook.md#alert-name-kubeclienterrors
+        expr: sum(rate(ksm_scrape_error_total{job="kube-state-metrics"}[5m])) by (instance, job) > 0.1
+        for: 15m
+        labels:
+          severity: warning
+      - alert: KubeletTooManyPods
+        annotations:
+          message: Kubelet {{`{{ $labels.instance }}`}} is running {{`{{ $value }}`}} Pods, close to the limit of 110.
```
damacus, line 117: Is the 110 a hard limit for Kubelets?

charts/prometheus-resources/values.yaml
```diff
@@ -72,3 +72,103 @@ prometheusRules:
...
+    - name: kubernetes-system
+      rules:
+      - alert: KubeNodeNotReady
+        annotations:
+          message: '{{`{{ $labels.node }}`}} has been unready for more than 5 minutes.'
+          runbook_url: https://github.com/kubernetes-monitoring/kubernetes-mixin/tree/master/runbook.md#alert-name-kubenodenotready
+        expr: kube_node_status_condition{job="kube-state-metrics",condition="Ready",status="true"} == 0
+        for: 5m
+        labels:
+          severity: warning
```
andylightfoot, line 86: Maybe up the severity to critical?

charts/prometheus-resources/values.yaml
andylightfoot: I think 110 is the default for Max pods per node.

charts/prometheus-resources/values.yaml
tomskelley-gjs: Fixed - OK to merge and apply?

PR #443, opened 2020-07-07
Updating helm charts for airflow

isichei:


charts/airflow-k8s/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [0.3.3] - 2020-07-07
```
xoen, line 7: Let's follow semantic versioning please ☺️ https://semver.org/

AFAICT this is adding features in a retro-compatible way (fingers crossed 🤞) so unless I'm missing something it should be bumped to `0.4.0`.

(BTW thanks for updating the Changelog as well 🌟 )

charts/airflow-k8s/Chart.yaml
```diff
@@ -2,5 +2,5 @@ apiVersion: v1
 description: Airflow on kubernetes. Tasks are run as pods
 name: airflow-k8s
 icon: https://airflow.incubator.apache.org/_images/pin_large.png
-version: 0.3.2
-appVersion: 1.10.6-1
+version: 0.3.3
+appVersion: 1.10.10
```
xoen, line 6: This is less important in this helm chart (as it's not user-facing) but we should move to something more detailed here (and more importantly in the `airflow-sqlite` chart), [something similar to what we have in the RStudio chart](https://github.com/ministryofjustice/analytics-platform-helm-charts/blob/master/charts/rstudio/Chart.yaml#L5).

Something like `Airflow: 1.10.10, Python: 3.7` (or whatever version of Python is shipped in this Docker image).

charts/airflow-sqlite/Chart.yaml
```diff
@@ -2,5 +2,5 @@ apiVersion: v1
 description: Airflow with sqlite. Tasks are run as pods
 name: airflow-sqlite
 icon: https://airflow.incubator.apache.org/_images/pin_large.png
-version: 0.3.2
-appVersion: 1.10.6-1
+version: 0.3.3
+appVersion: 1.10.10
```
xoen, line 6: Basically same comment I made in main airflow chart: https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/443/files#r450897241

In this chart it's even more important as the user will be able to see in the CP what things are shipped in the tool and if there is anything they should be aware of (e.g. moving from Python 3.6 to Python 4.1, where versions numbers are made up but you get the point)

charts/airflow-k8s/Chart.yaml
isichei: Ah I see - I get you now! Will do

PR #445, opened 2020-07-10
Add reset user home chart

tomskelley-gjs:
https://trello.com/c/CxKqlUpN/654-ripley-button-resets-a-users-conda-environment-etc

### DIFF
--- a/charts/reset-user-home/Chart.yaml
+++ b/charts/reset-user-home/Chart.yaml
@@ -1,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
+description: Reset user home environment
+name: reset-user-home
+version: 0.1.0
--- a/charts/reset-user-home/values.yaml
+++ b/charts/reset-user-home/values.yaml
@@ -1,0 +1,23 @@
+# Default values for reset-user-home.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+username: ""
+
+replicaCount: 1
+
+image:
+  repository: "quay.io/mojanalytics/airflow"
+  tag: "1.10.6-1"
+  pullPolicy: "IfNotPresent"
+  
+nameOverride: ""
+fullnameOverride: ""
+  
+resources: {}
+  
+nodeSelector: {}
+
+tolerations: []
+
+affinity: {}
--- a/charts/reset-user-home/.helmignore
+++ b/charts/reset-user-home/.helmignore
@@ -1,0 +1,22 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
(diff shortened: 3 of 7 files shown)


damacus:
I'm missing something with the design here (possibly?), what's the use of the Ingress controller here?

charts/reset-user-home/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
+description: Rest user home environment
```
andylightfoot, line 3: Small typo with Rest vs Reset

charts/reset-user-home/templates/ingress.yaml
```diff
@@ -0,0 +1,39 @@
+{{- if .Values.ingress.enabled -}}
```
andylightfoot, line 1: I don't think you need an ingress for this

charts/reset-user-home/templates/service.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
```
andylightfoot, line 1: Not sure you need a service for this either

charts/reset-user-home/values.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+image:
+  repository: "quay.io/mojanalytics/airflow"
+  tag: "1.10.6-1"
+  pullPolicy: "IfNotPresent"
+  
+nameOverride: ""
+fullnameOverride: ""
+
+service:
```
andylightfoot, line 17: Not sure this needs a service?

charts/reset-user-home/values.yaml
```diff
@@ -0,0 +1,37 @@
...
+  type: ClusterIP
+  port: 80
+
+ingress:
```
andylightfoot, line 21: Not sure this needs an ingress?

charts/reset-user-home/Chart.yaml
tomskelley-gjs: Fixed

charts/reset-user-home/templates/ingress.yaml
tomskelley-gjs: Removed - was added by the template - I just didn't remove it

charts/reset-user-home/templates/service.yaml
tomskelley-gjs: Removed - was also added by the template - I just didn't remove it

charts/reset-user-home/values.yaml
tomskelley-gjs: Removed

charts/reset-user-home/values.yaml
tomskelley-gjs: Removed

PR #447, opened 2020-07-13
Add concourse helm chart

andylightfoot:
Added the concourse chart as a copy of the stable/concourse chart as we need to make some small changes to it to improve the stability of the concourse workers.

These changes are not configurable in the current version of the concourse chart and we don't want to upgrade concourse at the moment as there are breaking changes in the later versions.

In the longer term we will revert to the stable/concourse chart when we upgrade.


andylightfoot:
After this has been merged, I'll raise another PR with the changes in.

PR #448, opened 2020-07-13
Add priority to concourse workers and remove old loop devices

andylightfoot:
We have seen issues with the Concourse workers pods failing on a regular basis due to 

- the PVCs that they use to cache config filling up
- workers failing to retire cleanly

The reason that the PVCs seem to be filling up is due to old loop devices being left around after a pods is restarted. As the worker always creates a new loop devices each time it start, we can delete all loop devices on startup which should ensure that old ones are not left around.

It isn't clear why the concourse workers are sometimes failing to retire correctly after restarting. However, if we can limit the number of times the pods are restarted it should reduce the number of time we see this issue. One way of use doing this is to increase the priority of the worker pods so that they are less likely to be rescheduled to a new Kubernetes node.


damacus:
[Update datascience-notebook to 229c7fea9d60](https://trello.com/c/m5LGzt2P/687-update-datascience-notebook-to-229c7fea9d60)

PR #450, opened 2020-07-13
 Trello 654 - Ripley Button Remove connectivity test 

tomskelley-gjs:
...as this is a job without an ingress so it's meaningless.

Trello 654 - Ripley Button


tomskelley-gjs:
This one got missed when I removed all the ingress etc stuff before.

PR #454, opened 2020-07-14
Append a 5 character random string to the job name.

tomskelley-gjs:
The correct way to do this is probably have a hook to delete the job

after it completes, per this:

https://github.com/helm/helm/issues/1769

But it seems bugged in multiple versions of helm, so this workaround will do for the time being.

### DIFF
--- a/charts/reset-user-home/Chart.yaml
+++ b/charts/reset-user-home/Chart.yaml
@@ -2,4 +2,4 @@ apiVersion: v1
 appVersion: "1.0"
 description: Reset user home environment
 name: reset-user-home
-version: 0.4.0
+version: 0.5.0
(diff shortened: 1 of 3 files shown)


charts/reset-user-home/Chart.yaml
```diff
@@ -2,4 +2,4 @@ apiVersion: v1
 appVersion: "1.0"
 description: Reset user home environment
 name: reset-user-home
-version: 0.4.0
+version: 0.5.0
```
damacus, line 5:
```suggestion
version: 0.4.1
```

damacus: Probably only needs to be a bug fix release

PR #456, opened 2020-07-17
692 investigate softnas backup

tomskelley-gjs:
Use *much* more recent AWS provided docker container.


charts/nfs-backup/templates/cronjob.yml
```diff
@@ -50,4 +95,4 @@ spec:
               claimName: nfs-homes
           - name: exclude-patterns
             configMap:
-              name: {{ template "fullname" . }}
+              name: {{ template "fullname" . }}
```
damacus, line 98: why no EOL character?

tomskelley-gjs: Doesn't need one. SSDs aren't cheap!

charts/nfs-backup/values.yaml
```diff
@@ -1,10 +1,10 @@
...
-  IAMRole: ""
+  IAMRole: "alpha_nfs_backup"
   S3:
-    BucketName: ""
+    BucketName: "alpha-moj-analytics-nfs-backup"
 Image:
-  Repository: samepagelabs/s3cmd
-  Tag: "1.6.1"
+  Repository: amazon/aws-cli
+  Tag: "latest"
```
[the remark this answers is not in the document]
tomskelley-gjs: Pinned

charts/nfs-backup/templates/cronjob.yml
tomskelley-gjs: Fixed just in case it breaks any automation

PR #457, opened 2020-07-22
Add a Prometheus rule for KubeDNS errors

andylightfoot:
Add a prometheus rule that will trigger an alert if the value of kubedns_dnsmasq_errors increases.

This should allow us to detect when KubeDNS pods log errors.

(Also correct annoying spelling mistake)

### DIFF
--- a/charts/prometheus-resources/Chart.yaml
+++ b/charts/prometheus-resources/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v1
 appVersion: "1.0"
 description: Deploys resources related to the [Prometheus-Operator](https://github.com/helm/charts/tree/master/stable/prometheus-operator) helm chart
 name: prometheus-resources
-version: 0.1.6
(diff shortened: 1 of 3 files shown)


charts/prometheus-resources/Chart.yaml
```diff
@@ -2,5 +2,5 @@ apiVersion: v1
 appVersion: "1.0"
 description: Deploys resources related to the [Prometheus-Operator](https://github.com/helm/charts/tree/master/stable/prometheus-operator) helm chart
 name: prometheus-resources
-version: 0.1.6
+version: 0.1.7
```
xoen, line 5: You may want to bump to `0.2.0` as you're adding a functionality in a retro-compatible way: https://semver.org/ (in addition to fixing a typo/bug)

andylightfoot: That's a nice idea, will do that.

PR #460, opened 2020-08-10
Add helm chart to apply limit ranges to kubernetes namespaces

andylightfoot:
If pods do not specify limits for CPU and Memory then can potentially use all the CPU and Memory on a node which can cause problems for other apps running on the same node.

We can apply Limit ranges to a namespace which will apply CPU and Memory limits and requests to containers in a pod if they don't specify them.

https://kubernetes.io/docs/concepts/policy/limit-range

### DIFF
--- a/charts/limit-ranges/Chart.yaml
+++ b/charts/limit-ranges/Chart.yaml
@@ -1,0 +1,7 @@
+apiVersion: v1
+description: Deploy Limit Ranges for Kubernetes namespaces
+name: limit-ranges
+version: 0.0.1
+sources:
+  - https://github.com/ministryofjustice/analytics-platform-helm-charts
+  - https://kubernetes.io/docs/concepts/policy/limit-range
--- a/charts/limit-ranges/README.md
+++ b/charts/limit-ranges/README.md
@@ -1,0 +1,33 @@
+# Limit ranges
+
+Creates `limit ranges` for a namespace [Limit Ranges](https://kubernetes.io/docs/concepts/policy/limit-range)
+
+### Installing
+
+```bash
+helm upgrade limit-ranges mojanalytics/limit-ranges --install
+```
+
+### Configuration
+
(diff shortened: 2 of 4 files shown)


damacus:
I wonder if we should look at doing this for things in the user namespaces too?

andylightfoot:
> I wonder if we should look at doing this for things in the user namespaces too?

Yeah we could do. We could add something to the user init process so that we also add limits requests.

xoen:
Hang on, actually [from the k8s documentation](https://kubernetes.io/docs/concepts/policy/limit-range/) seems that this will force things to not use more than the limits:

> A LimitRange is a policy to constrain resource allocations (to Pods or Containers) in a namespace.
> [...]
> Enforce minimum and maximum compute resources usage per Pod or Container in a namespace.

👌 

I also notice that: 

> If creating or updating a resource (Pod, Container, PersistentVolumeClaim) that violates a LimitRange constraint, the request to the API server will fail with an HTTP status code 403 FORBIDDEN and a message explaining the constraint that have been violated.

Hopefully this will not break anything 🤞

andylightfoot:
> from the k8s documentation

It is a good point Aldo and I had to re-read the Kubernetes docs a couple of times to understand limit ranges fully.

It looks like you can create different type of limit ranges, default, default requests, min, max

I think we should be ok because we've just created a default Limit range 

i.e. https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/cpu-default-namespace/

instead of a max/min limits 
i.e.
https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/memory-constraint-namespace/

xoen:
Oh I see now! So we should be fine as you pointed out as this is just a default in case apps' pods didn't have any limit 👌 

We'll probably have to tweak the values' structure later on when/**if** we'll use min/max (to distinguish between the default and other constraints) but for now this is a great start ⭐

PR #466, opened 2020-09-04
cpanel 2.4.3: reverse-proxy container readinessProbe

xoen:
Using newly added `/nginx-health` endpoint in the CP nginx config to
periodically check that nginx is still alive and serving traffic.

See: https://github.com/ministryofjustice/analytics-platform-control-panel/pull/834/commits/d2e2ba33f46205b66f3060009bbc490890e52275
Part of: https://trello.com/c/eUMRw1gK

### DIFF
--- a/charts/cpanel/templates/deployment.yaml
+++ b/charts/cpanel/templates/deployment.yaml
@@ -51,6 +51,12 @@ spec:
               mountPath: /etc/nginx/conf.d
             - name: staticfiles
               mountPath: /usr/share/nginx/html
+          readinessProbe:
+            httpGet:
+              path: /nginx-health
+              port: http
+            initialDelaySeconds: 5
+            periodSeconds: 5
         - name: worker
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           imagePullPolicy: "{{ .Values.image.pullPolicy }}"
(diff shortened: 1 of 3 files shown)


charts/cpanel/templates/deployment.yaml
```diff
@@ -51,6 +51,12 @@ spec:
               mountPath: /etc/nginx/conf.d
             - name: staticfiles
               mountPath: /usr/share/nginx/html
+          readinessProbe:
+            httpGet:
+              path: /nginx-health
+              port: http
+            initialDelaySeconds: 5
+            periodSeconds: 5
```
andylightfoot, line 59: Do you want to increase the failureThreshold for this or is the default value (3)ok?

andylightfoot: actually as this is nginx it should be fine!

andylightfoot (approved):
LGTM

charts/cpanel/templates/deployment.yaml
xoen: Yeah I think the default is fine for nginx, I don't expect this to fail much or at all TBH, especially as we don't have a huge amount of traffic anyway. And this is readiness probe rather than liveliness probe so it should be OK, but I guess we can tweak later on if we notice this is not OK anymore.

xoen:
Thanks @andylightfoot ❤️

PR #467, opened 2020-09-04
Add GitHub Actions runner chart

damacus:
- Adds Helm Chart for GitHub Actions runner
- Adds example values.yaml
- Adds a deployment for the single container pod
- Adds a secret for the GitHub Organisation token

### DIFF
--- a/charts/github-actions-runner/values.yaml
+++ b/charts/github-actions-runner/values.yaml
@@ -1,0 +1,22 @@
+---
+image:
+  repository: myoung34/github-runner
+  pullPolicy: Always
+  tag: "2.273.0"
+
+env:
+  org:
+    runner: "true"
+    name: ""
+    secret:
+      access_token:
+        name: github-token
+  labels: "kubernetes"
+
+autoscaling:
+  enabled: true
+  minReplicas: 1
+  maxReplicas: 50
+  targetCPUUtilizationPercentage: 80
+
+GitHubToken: ""
--- a/charts/github-actions-runner/.helmignore
+++ b/charts/github-actions-runner/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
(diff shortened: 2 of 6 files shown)


charts/github-actions-runner/values.yaml
```diff
@@ -0,0 +1,22 @@
+---
+image:
+  repository: myoung34/github-runner
```
andylightfoot, line 3: I'm not sure I'm that keen on us using docker images from personal repos for our github runners

charts/github-actions-runner/values.yaml
```diff
@@ -0,0 +1,22 @@
...
+env:
+  org:
+    runner: "true"
+    name: ""
+    secret:
+      access_token:
+        name: github-token
+  labels: "kubernetes"
+
+autoscaling:
```
andylightfoot, line 16: How does the autoscaling work?

charts/github-actions-runner/values.yaml
damacus: Granted. Though, this looks to be the best community supports Docker image. 

We could look at cloning and building our own, but that seems like a lot of extra work for nebulous gains.

charts/github-actions-runner/values.yaml
damacus: I'm fairly sure it works with CPU "pressure". Happy to move to a fixed of runners if this doesn't work in heavy testing.

charts/github-actions-runner/values.yaml
andylightfoot: These runners will have potentially very powerful IAM permissions in our AWS account so I think it is wrong to discuss concerns about the security of this image as nebulous.

charts/github-actions-runner/values.yaml
andylightfoot: Yeah but what is actually scaling it based on CPU Pressure? I can't see it adding a Horizontal Pod Autoscaler

charts/github-actions-runner/values.yaml
andylightfoot: I don't think this should block the POC of github runners but lets review and come to a consensus on this before giving too many IAM perms

andylightfoot (approved):
Maybe go with fixed number of replicas for now then work out autoscaling later on?

PR #468, opened 2020-09-11
cpanel 2.4.4: Added liveliness probe to `worker` (Django Channels)

xoen:
### What

Use newly added CP command to check whether a worker is still processing
background tasks at a reasonable rate or it's somehow stuck.

### How
This is achieved by using the CP `worker_health` command (see PR below)

```
$ python3 manage.py worker_health --stale-after-secs=120
```

which will periodically submit health tasks and check if one has been
performed recently (withing 2 minutes in this case, configurable using
`----stale-after-secs=120`)

Probes are sent every 30s but note how one or more individual probes can
fail without triggering the restart of the `worker` container.
This is by design as you don't want Kubernetes to be happy-trigger and
restart this container too easily.

`successThreshold=6` means that 6 *consecutive* probes will have to fail
in order to trigger a container restart (that is nothing was run for few
minutes on that specific pod/worker container).

Also note how in reality tasks will be performed by each worker at fairly
regular intervals of every ~30s (confirmed by testing in `dev` environment)

### Resources
Also see Control Panel command PR: ministryofjustice/analytics-platform-control-panel#840

Part of ticket: https://trello.com/c/eUMRw1gK

### DIFF
--- a/charts/cpanel/templates/deployment.yaml
+++ b/charts/cpanel/templates/deployment.yaml
@@ -68,6 +68,18 @@ spec:
               name: "{{ .Chart.Name }}-env-secrets"
           - secretRef:
               name: "{{ .Chart.Name }}-db-env"
+          livenessProbe:
+              exec:
+                command:
+                  - python3
+                  - manage.py
+                  - worker_health
+                  - "--stale-after-secs=120"
+              initialDelaySeconds: 10
+              periodSeconds: 30
+              failureThreshold: 6
+              successThreshold: 1
+              timeoutSeconds: 30
         - name: backend
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           imagePullPolicy: "{{ .Values.image.pullPolicy }}"
--- a/charts/cpanel/CHANGELOG.md
+++ b/charts/cpanel/CHANGELOG.md
@@ -5,6 +5,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+## [2.4.4] - 2020-09-11
+### Added
+- Added liveliness probe to `worker` (Django Channels) container
+
+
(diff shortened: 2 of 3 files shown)


xoen:
FYI: I haven't merged [this other probes-related PR]() so this diff also contains those changes but [the commit for this PR you want to look at (diff-wise) is really this](https://github.com/ministryofjustice/analytics-platform-helm-charts/commit/ce11c76b8f28e0ebe56fb48e9a94c3c8aa437ad5).

charts/cpanel/templates/deployment.yaml
```diff
@@ -62,6 +68,18 @@ spec:
               name: "{{ .Chart.Name }}-env-secrets"
           - secretRef:
               name: "{{ .Chart.Name }}-db-env"
+          livenessProbe:
+              exec:
+                command:
+                  - python3
```
andylightfoot, line 74: V minor point, but I think these should be aligned directing under the command: (i.e. move left 2 chars)

andylightfoot (approved):
I think maybe the commands need realigning slightly but it is YAML so maybe it doesn't matter?

charts/cpanel/templates/deployment.yaml
xoen: AFAICT Yaml lists works both ways but I always find the indentation useful as it makes it much more clear that these are elements of the list.

Having said that, le me try again with this YAML and see if it works as I think I do - I think I did it on Friday but let me triple check 👍

xoen: FYI: Yeah I've double-checked and it's valid YAML and also helps with visualising that these strings are part of the same list rather than at the same level of `command` and the rest so I'll keep the indentation unless there are major objections (which I don't think there are).

xoen:
Thanks for the review Andy ❤️

charts/cpanel/templates/deployment.yaml
andylightfoot: Great, happy to go with that.

PR #469, opened 2020-09-14
cpanel 2.4.5: tweaks to worker livelinessProbe

xoen:
Further tweaking liveliness probe for `worker` (Django Channels) container
to account for potentially very slow background tasks blocking a specific
worker.

Specifically new Deploy/Initialisation of RStudio can be VERY slow
and CP is doing a `helm upgrade --wait` so it is blocking the worker:

https://github.com/ministryofjustice/analytics-platform-control-panel/blob/main/controlpanel/frontend/consumers.py#L109-L132

The default `--timeout` for `helm` is 300s/5 minutes:

https://helm.sh/docs/helm/helm_upgrade/#options

So I'm tweaking the worker liveliness probe to consider a worker stalled when
it didn't run something for 5 minutes and reduce the chance that a worker
container is restarted just because someone installed a tool.

Part of ticket: https://trello.com/c/eUMRw1gK

### DIFF
--- a/charts/cpanel/CHANGELOG.md
+++ b/charts/cpanel/CHANGELOG.md
@@ -5,6 +5,26 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+## [2.4.5] - 2020-09-14
+### Changed
+Further tweaking liveliness probe for `worker` (Django Channels) container
+to account for potentially very slow background tasks blocking a specific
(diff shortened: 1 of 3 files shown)


xoen:
Sorry @andylightfoot was doing some more manual testing in `dev` and realised the probe could use some tweaking.

I've explained why in PR/changelog but let me know if something is unclear. Effectively I'm slightly increasing the acceptable amount of time a worker hasn't done work for to consider it "stuck".

andylightfoot (approved):
LGTM

xoen:
Thanks again!

PR #470, opened 2020-09-14
cpanel 2.4.6: Moved `worker` in separate pod

xoen:
The `worker` container is decoupled from the nginx/django frontend containers
and there is no reason for it to be in the same pod.

The workers are decoupled as they receive background jobs via redis.

Furthermore, have it in a separate `Deployment` means we can scale workers
independently from the frontend.

This also increase the number of workers replicas to `10` from `3`
(which was the number of frontend replicas).

I've added a `frontend.replicas` and `worker.replicas` values so that
these can potentially be tweaked/configured more easily.

Ticket: https://trello.com/c/Bbl5R56H


xoen:
FYI: I think [this helm comment](https://github.com/helm/helm/issues/2861#issuecomment-328096540) basically confirm what I thought about the existing [`cpanel` `ClusterRole`](https://github.com/ministryofjustice/analytics-platform-helm-charts/blob/da106868da45ee68bfa15769f1497b31681592e8/charts/cpanel/templates/cluster-role.yaml#L1) (which I've now renamed to `cpanel-helm-client` for clarity): These permissions were there to allow the client to find/talk with tiller (helm's server).

charts/cpanel/templates/worker-deployment.yaml
```diff
@@ -0,0 +1,54 @@
...
+          livenessProbe:
+              exec:
+                command:
+                  - python3
+                  - manage.py
+                  - worker_health
+                  - "--stale-after-secs=300"
+              initialDelaySeconds: 10
+              periodSeconds: 30
+              failureThreshold: 6
```
andylightfoot, line 52: If we have 10 workers then is the failureThreshold of 6 high enough?

andylightfoot (approved):
A great piece of work Aldo! LGTM.

charts/cpanel/templates/worker-deployment.yaml
xoen: AFAICT from previous testing in `dev` it shouldn't matter (**but I'll test**).

The reason being that workers seem to process background jobs very uniformly (at least when not blocked installing RStudio or similar). So every `30s` k8s will send a probe into each of the N workers containers, which will submit the job.

That means that every `~30s` N probes will be sent and N jobs will be processed by the workers so **on average**, each worker process roughly 1 job every `~30s`.

The `6` here mainly determine how long a worker wouldn't have not performed anything before we consider it stuck.
If we increase this k8s will take longer to take action when workers will be stuck for real.

It's a bit tricky to visualise but imagine `6` (currently) consecutive probes each checking if the worker did anything in the last `5m` (`--stale-after-secs=300`), each sent every `30s`, this means (unless I'm mistaken) that at maximum a worker could be stuck for `5m` + `6`x `30s` before it is restarted (`8m` = `5`+`3m`).

Increasing to `10` as you said would make k8s less happy trigger but it would also increase the amount of time a worker may be stuck without being restarted, e.g. `10m` (`5m` + `5m` = `5m` + `10` x `0.5m`).

In [this OpenShit article on Liveness and Readiness probes](https://www.openshift.com/blog/liveness-and-readiness-probes) they have a section and some graphs on probes timing, etc which may help think about this. 

Of course from testing we may realise we need to tweak these probes further, but luckily everything is configurable ☺️

PR #471, opened 2020-09-15
reset-user-home 0.4.2: Deeper Conda cleanup

xoen:
Archiving not only the `~/.conda/envs/rstudio` environment but also the
whole `~/.conda/` directory AND the `~/.condarc` config file.

This is to address RStudio startup problems caused by other conda
environments and/or broken conda configuration on JupyterLab (or whatever).

Ticket: https://trello.com/c/TWlH6wt1

### DIFF
--- a/charts/reset-user-home/CHANGELOG.md
+++ b/charts/reset-user-home/CHANGELOG.md
@@ -1,0 +1,19 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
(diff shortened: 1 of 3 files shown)


andylightfoot (approved):
Nice!

ntoll (approved):
LGTM...! Delete all the things...!

xoen:
FYI: Tested this in `dev` in own namespace/home directory and it worked  (see below). As it seems like there are no objections I'm going to merge this.

(to test I've exec into a container and created a new `~/.condarc` file and a new directory under `~/.conda/envs/` and installed the helm chart from this branch. I've also tested the case when these were not already present and re-installed this helm chart and it worked as well as expected)

PR #474, opened 2020-10-01
cpanel: Removed helm dependencies to PostgreSQL and Redis

xoen:
### What

Both shouldn't run on the cluster and they should run on RDS or ElastiCache. In fact:
- https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/474/commits/7127ad2274898787ad1227e4de0f975e575969ae We **already** don't deploy/use this k8s PostgreSQL as CP DB runs on AWS RDS (who wants to manage/maintain DBs?) so this dependency is unnecessary and confusing (nice to be able to removed a bunch of unused helm helpers which are not used and confusing)
-  https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/474/commits/a0d8200396ca6109a6003dfe38982fd23e73cfeb We're in the process of pointing to a Redis cluster hosted on AWS EC, which is why we're making these changes

### Breaking Changes
This [latest version of `cpanel` helm chart](https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/474/commits/a0d8200396ca6109a6003dfe38982fd23e73cfeb) has breaking changes:
- the `redis` helm chart dependency is removed which means Redis will need to be available somewhere else at the time of deploy
- the `redis.host` value will need to be provided so that CP knows where this other Redis cluster is
- the `redis.redisPort` is been renamed to just `redis.port` for consistency. We always used the default port so it shouldn't really break but technically it's a breaking change to be aware of (in case someone else out there is using this helm chart somehow)

### Ticket
https://trello.com/c/YG9anK8A


damacus (approved):

andylightfoot (approved):
LGTM

PR #484, opened 2020-10-30
Use mv rather than copy to backup old directories when resetting user's home

ntoll:
* Use `mv -f` to force the rename.
* I've removed the superfluous `rm -rf` of the path that's just been `mv`-ed.
* No idea how to test this. #helmnewbie ;-)

Please feel free to ignore.

Related Trello card: https://trello.com/c/eEmNEwI7/830-make-home-reset-faster

### DIFF
--- a/charts/reset-user-home/templates/deployment.yaml
+++ b/charts/reset-user-home/templates/deployment.yaml
@@ -18,13 +18,13 @@ spec:
             - /bin/sh
             - -c
             - |
-              if test -d /home/{{ .Values.Username }}/.rstudio.bak; then rm -fr /home/{{ .Values.Username }}/.rstudio.bak; fi &&
-              if test -d /home/{{ .Values.Username }}/.rstudio; then rm -rf /home/{{ .Values.Username }}/.rstudio /home/{{ .Values.Username }}/.rstudio.bak; fi &&
-              if test -d /home/{{ .Values.Username }}/.conda.bak/; then rm -fr /home/{{ .Values.Username }}/.conda.bak/; fi &&
-              if test -d /home/{{ .Values.Username }}/.conda/; then rm -rf /home/{{ .Values.Username }}/.conda/ /home/{{ .Values.Username }}/.conda.bak/; fi &&
-              if test -e /home/{{ .Values.Username }}/.condarc.bak; then rm -fr /home/{{ .Values.Username }}/.condarc.bak; fi &&
-              if test -e /home/{{ .Values.Username }}/.condarc; then rm -rf /home/{{ .Values.Username }}/.condarc /home/{{ .Values.Username }}/.condarc.bak; fi &&
-              if test -d /home/{{ .Values.Username }}/R/library; then rm -fr /home/{{ .Values.Username }}/R/library/*; fi
+              rm -rf /home/{{ .Values.Username }}/.rstudio.bak; &&
+              rm -rf /home/{{ .Values.Username }}/.rstudio; &&
+              rm -rf /home/{{ .Values.Username }}/.conda.bak/; &&
+              rm -rf /home/{{ .Values.Username }}/.conda/; &&
+              rm -rf /home/{{ .Values.Username }}/.condarc.bak; &&
+              rm -rf /home/{{ .Values.Username }}/.condarc;  &&
+              rm -rf /home/{{ .Values.Username }}/R/library/*;
       volumes:
         - name: home
           persistentVolumeClaim:
--- a/charts/reset-user-home/templates/deployment.yaml
+++ b/charts/reset-user-home/templates/deployment.yaml
@@ -21,7 +21,7 @@ spec:
               if test -d /home/{{ .Values.Username }}/.rstudio.bak; then rm -fr /home/{{ .Values.Username }}/.rstudio.bak; fi &&
               if test -d /home/{{ .Values.Username }}/.rstudio; then rm -rf /home/{{ .Values.Username }}/.rstudio /home/{{ .Values.Username }}/.rstudio.bak; fi &&
               if test -d /home/{{ .Values.Username }}/.conda.bak/; then rm -fr /home/{{ .Values.Username }}/.conda.bak/; fi &&
-              if test -d /home/{{ .Values.Username }}/.conda/; then mv -f /home/{{ .Values.Username }}/.conda/ /home/{{ .Values.Username }}/.conda.bak/; fi &&
+              if test -d /home/{{ .Values.Username }}/.conda/; then rm -rf /home/{{ .Values.Username }}/.conda/ /home/{{ .Values.Username }}/.conda.bak/; fi &&
               if test -e /home/{{ .Values.Username }}/.condarc.bak; then rm -fr /home/{{ .Values.Username }}/.condarc.bak; fi &&
               if test -e /home/{{ .Values.Username }}/.condarc; then mv -f /home/{{ .Values.Username }}/.condarc /home/{{ .Values.Username }}/.condarc.bak; fi &&
               if test -d /home/{{ .Values.Username }}/R/library; then rm -fr /home/{{ .Values.Username }}/R/library/*; fi
--- a/charts/reset-user-home/templates/deployment.yaml
+++ b/charts/reset-user-home/templates/deployment.yaml
@@ -23,7 +23,7 @@ spec:
               if test -d /home/{{ .Values.Username }}/.conda.bak/; then rm -fr /home/{{ .Values.Username }}/.conda.bak/; fi &&
               if test -d /home/{{ .Values.Username }}/.conda/; then rm -rf /home/{{ .Values.Username }}/.conda/ /home/{{ .Values.Username }}/.conda.bak/; fi &&
               if test -e /home/{{ .Values.Username }}/.condarc.bak; then rm -fr /home/{{ .Values.Username }}/.condarc.bak; fi &&
-              if test -e /home/{{ .Values.Username }}/.condarc; then mv -f /home/{{ .Values.Username }}/.condarc /home/{{ .Values.Username }}/.condarc.bak; fi &&
+              if test -e /home/{{ .Values.Username }}/.condarc; then rm -rf /home/{{ .Values.Username }}/.condarc /home/{{ .Values.Username }}/.condarc.bak; fi &&
               if test -d /home/{{ .Values.Username }}/R/library; then rm -fr /home/{{ .Values.Username }}/R/library/*; fi
       volumes:
         - name: home
--- a/charts/reset-user-home/templates/deployment.yaml
+++ b/charts/reset-user-home/templates/deployment.yaml
@@ -19,7 +19,7 @@ spec:
             - -c
             - |
               if test -d /home/{{ .Values.Username }}/.rstudio.bak; then rm -fr /home/{{ .Values.Username }}/.rstudio.bak; fi &&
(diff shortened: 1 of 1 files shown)


damacus (approved):

xoen (approved):
This looks good to me 👍 

Only thing will have to bear in mind in case we'll have to restore any of these `.bak` directories/files is that `mv` doesn't preserve the permissions/ownership while `cp -p` did.

👆 BUT, I don't think this is a huge problem or a blocker, if we'll ever have to restore these we can always adjust the permissions.

charts/reset-user-home/templates/deployment.yaml
```diff
@@ -19,14 +19,11 @@ spec:
             - -c
             - |
               if test -d /home/{{ .Values.Username }}/.rstudio.bak; then rm -fr /home/{{ .Values.Username }}/.rstudio.bak; fi &&
-              if test -d /home/{{ .Values.Username }}/.rstudio; then cp -pRf /home/{{ .Values.Username }}/.rstudio /home/{{ .Values.Username }}/.rstudio.bak; fi &&
-              if test -d /home/{{ .Values.Username }}/.rstudio; then rm -fr /home/{{ .Values.Username }}/.rstudio; fi &&
+              if test -d /home/{{ .Values.Username }}/.rstudio; then mv -f /home/{{ .Values.Username }}/.rstudio /home/{{ .Values.Username }}/.rstudio.bak; fi &&
```
damacus, line 22:
```suggestion
              if test -d /home/{{ .Values.Username }}/.rstudio; then rm -rf /home/{{ .Values.Username }}/.rstudio /home/{{ .Values.Username }}/.rstudio.bak; fi &&
```

charts/reset-user-home/templates/deployment.yaml
```diff
@@ -19,14 +19,11 @@ spec:
...
               if test -d /home/{{ .Values.Username }}/.conda.bak/; then rm -fr /home/{{ .Values.Username }}/.conda.bak/; fi &&
-              if test -d /home/{{ .Values.Username }}/.conda/; then cp -pRf /home/{{ .Values.Username }}/.conda/ /home/{{ .Values.Username }}/.conda.bak/; fi &&
-              if test -d /home/{{ .Values.Username }}/.conda/; then rm -fr /home/{{ .Values.Username }}/.conda/; fi &&
+              if test -d /home/{{ .Values.Username }}/.conda/; then mv -f /home/{{ .Values.Username }}/.conda/ /home/{{ .Values.Username }}/.conda.bak/; fi &&
```
damacus, line 24:
```suggestion
              if test -d /home/{{ .Values.Username }}/.conda/; then rm -rf /home/{{ .Values.Username }}/.conda/ /home/{{ .Values.Username }}/.conda.bak/; fi &&
```

charts/reset-user-home/templates/deployment.yaml
```diff
@@ -19,14 +19,11 @@ spec:
...
               if test -e /home/{{ .Values.Username }}/.condarc.bak; then rm -fr /home/{{ .Values.Username }}/.condarc.bak; fi &&
-              if test -e /home/{{ .Values.Username }}/.condarc; then cp -pRf /home/{{ .Values.Username }}/.condarc /home/{{ .Values.Username }}/.condarc.bak; fi &&
-              if test -e /home/{{ .Values.Username }}/.condarc; then rm -fr /home/{{ .Values.Username }}/.condarc; fi &&
+              if test -e /home/{{ .Values.Username }}/.condarc; then mv -f /home/{{ .Values.Username }}/.condarc /home/{{ .Values.Username }}/.condarc.bak; fi &&
```
damacus, line 26:
```suggestion
              if test -e /home/{{ .Values.Username }}/.condarc; then rm -rf /home/{{ .Values.Username }}/.condarc /home/{{ .Values.Username }}/.condarc.bak; fi &&
```

damacus (changes requested):

ntoll:
Waiting on performance test results from @andylightfoot :+1:

xoen:
@nicholsondarius / @andylightfoot / @damacus I've done some testing in my RStudio pod in `dev`.
TL;DR: on a 1.5GB `.conda/` directory  `cp` takes 3-4 minutes, `rm` takes 40-45s and `mv` is almost instantaneous. 

# cp vs. mv vs. rm

### `.conda` size

In `dev` my `.conda` directory with `rstudio` environment was 1.5GB

```sh
du -h /home/xoen/.conda/
[...]
1.5G	/home/xoen/.conda/
```

### Copy performances (cp -pRf)

Note: This copy will also keep the existing permissions/ownership

```sh
time cp -pRf /home/xoen/.conda/ /home/xoen/.conda.bak/

real	3m56.868s
user	0m0.952s
sys	0m16.740s
```

This will change, I've run this few times, sometimes it takes ~3m instead
of almost 4m but it's still slower than removing this directory with `rm`.

### rm of the copy performance

```sh
time rm -rf /home/xoen/.conda.bak/

real	0m41.325s
user	0m0.264s
sys	0m3.932s
```

`rm`ing a directory take ~83% LESS than copying its content (but again
time will vary).
Not instant but still significantly faster than a `cp`.

### Move performances (mv -f)

```sh
time mv /home/xoen/.conda/ /home/xoen/.conda.bak/

real	0m0.003s
user	0m0.000s
sys	0m0.000s
```

Move a directory seem to be instant. This is to be expected as it just mv
pointers to file rather than data.


(also it appears like `mv` kept the file permissions)

damacus:
@xoen I think ~45 seconds for a fix on this one seems pretty reasonable.

xoen:
@damacus yeah I'm more in favour of `rm` (FWIW) as we also reduce the disk usage instead of having stuff amassing in their home directory and effectively never cleaned (ATM)

Only thing I'm wondering is if some users will have even more than 1.5GB of rubbish in `.conda`/`.rstudio`/etc and it would take even longer than 45s...

I guess the question is now: would `rm` may still take too for some users? (despite being significantly faster)

I don't know what you think...

ntoll:
OK folks... since nobody has suggested otherwise, I'll commit @damacus 's updates and merge.

ntoll:
OK... so I can't merge until @damacus says I've made the requested changes. :-)

andylightfoot:
> @nicholsondarius / @andylightfoot / @damacus I've done some testing in my RStudio pod in `dev`.
> TL;DR: on a 1.5GB `.conda/` directory:
> 
> * `cp` takes 3-4 minutes
> * `rm` takes 40-45 seconds
> * `mv` is almost instantaneous.
> 
> # cp vs. mv vs. rm
> ### `.conda` size
> In `dev` my `.conda` directory with `rstudio` environment was 1.5GB
> 
> ```shell
> du -h /home/xoen/.conda/
> [...]
> 1.5G	/home/xoen/.conda/
> ```
> 
> ### Copy performances (cp -pRf)
> Note: This copy will also keep the existing permissions/ownership
> 
> ```shell
> time cp -pRf /home/xoen/.conda/ /home/xoen/.conda.bak/
> 
> real	3m56.868s
> user	0m0.952s
> sys	0m16.740s
> ```
> 
> This will change, I've run this few times, sometimes it takes ~3m instead
> of almost 4m but it's still slower than removing this directory with `rm`.
> 
> ### rm of the copy performance
> ```shell
> time rm -rf /home/xoen/.conda.bak/
> 
> real	0m41.325s
> user	0m0.264s
> sys	0m3.932s
> ```
> 
> `rm`ing a directory take ~83% LESS than copying its content (but again
> time will vary).
> Not instant but still significantly faster than a `cp`.
> 
> ### Move performances (mv -f)
> ```shell
> time mv /home/xoen/.conda/ /home/xoen/.conda.bak/
> 
> real	0m0.003s
> user	0m0.000s
> sys	0m0.000s
> ```
> 
> Move a directory seem to be instant. This is to be expected as it just mv
> pointers to file rather than data.
> 
> (also it appears like `mv` kept the file permissions)

Nice one Aldo. I also did a few tests in alpha with my .conda directory (~1.2GB) and the results were very similar to Aldo's.

damacus:
This all said, do we mind the `rm -rf` taking a minute if we reclaim the space? We're going to have to remove these files at some point anyway right?

ntoll:
@damacus that `rm -rf` takes *up to* a minute is a worst case scenario. In most cases it'll take a lot less. Also, we can update the message shown to users to manage their expectations: i.e. warning them that if they've conda'ed all the things, it may take a minute or two to complete. Personally, I vote +1 on `rm -rf` with a separate, but related, PR to update the messaging around the Ripley button for the users.

xoen:
@ntoll / @damacus I think deleting these directories/files is fair enough and, as I said, I like the idea of removing old stuff and free up some space as well 👍 

Hopefully extreme cases of users with GBs and GBs to remove are rare and will not take that much to delete anyway.

charts/reset-user-home/templates/deployment.yaml
```diff
@@ -19,14 +19,11 @@ spec:
             - -c
             - |
               if test -d /home/{{ .Values.Username }}/.rstudio.bak; then rm -fr /home/{{ .Values.Username }}/.rstudio.bak; fi &&
-              if test -d /home/{{ .Values.Username }}/.rstudio; then cp -pRf /home/{{ .Values.Username }}/.rstudio /home/{{ .Values.Username }}/.rstudio.bak; fi &&
-              if test -d /home/{{ .Values.Username }}/.rstudio; then rm -fr /home/{{ .Values.Username }}/.rstudio; fi &&
+              if test -d /home/{{ .Values.Username }}/.rstudio; then rm -rf /home/{{ .Values.Username }}/.rstudio /home/{{ .Values.Username }}/.rstudio.bak; fi &&
```
xoen, line 22: If we pass both `~/.rstudio/` and `~/.rstudio.bak/` as arguments to `rm`, we should be able to remove the line above which checks/delete the `~/.rstudio.bak/` directory.

Similarly, we can remove these lines which remove old backup in rest of the script.

I also wonder if we want to keep the script simpler and remove the logic related to the backups as it's not relevant anymore (although this may mean some users may still have some leftover backup in their homes)

ntoll:
OK simplified all the things. Please check..! Thanks for all the feedback folks. :+1:

charts/reset-user-home/templates/deployment.yaml
xoen: As said, the line above is redundant because of the space, e.g. this line will remove both `~/.rstudio/` AND `~/.rstudio.bak` so the one above is not needed.

Furthermore, the `test -d` can go I think because `rm -rf NOT-exists/ EXISTS/` will ignore a non-existent file/directory and delete the one which is there.

xoen (changes requested):
Maybe some commit was not pushed to the upstream repository yet.

I see the code is the same, e.g. has the `test -d`, and duplicated `rm` which are now not necessary ([see comment](https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/484/files#r522987247)).

Basically, probably the `test -d` can go and also some of the commands which are now covered by the modified `rm -rf`s which now remove 2 directories at the time.

ntoll:
@xoen dammit.... that's not the version of the code I have on my machine. *sigh* Lemme see what's going on. Probably my fault for not pushing something.

ntoll:
@xeon , so sorry about that... OK... my changes are now correct. Please try again. :-/

xoen (approved):
Awesome, **much** cleaner without all those checks 😍  

(I guess it could really be a huge single `rm -rf` but I think this way it's cleaner, easier to read and extend and also if any of them fails it's much more informative)

💯

xoen:
(@ntoll no problem!! I guessed some commit was not pushed to `origin` ☺️ )

damacus (approved):

PR #485, opened 2020-11-02
Tidy the runner configuration

damacus:
# Overview 

- Adds org_runner and repo_runner example configuration
- Namespace the runner now we also have dind configuration
- Make the dockerhost configurable
- Make the Docker image configurable
- Make runner groups configurable

## What

Adding org_runner and repo_runner configuration make it easier to test the logic in the deployment.yaml

Namespacing the runner values makes the hierarchy easier to understand for new users, and clarifies the shared values i.e. the values that are not under either the dind or runner key.

Making the DIND container image configurable makes it easy to update the Docker version without needing to update the Helm chart every time.

We can now configure runners to register with a specified runner group.

## Trello link
https://trello.com/c/ZNtuRljD/819-github-actions-runner

## Testing

Run

```shell
helm install runner . --namespace runners --dry-run -f org_runner.yaml
```

```shell
helm install runner . --namespace runners --dry-run -f repo_runner.yaml
```

### DIFF
--- a/charts/github-actions-runner/org_runner.yaml
+++ b/charts/github-actions-runner/org_runner.yaml
@@ -1,0 +1,5 @@
+---
+runner:
+  orgName: "foo"
+  labels: "kubernetes,foo,bar"
+  runnerGroup: "specialGroup"
--- a/charts/github-actions-runner/templates/deployment.yaml
+++ b/charts/github-actions-runner/templates/deployment.yaml
@@ -19,41 +19,43 @@ spec:
       labels:
         app.kubernetes.io/instance: {{ .Release.Name | quote }}
         app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
-      {{- if .Values.podAnnotations }}
+      {{- if .Values.runner.podAnnotations }}
       annotations:
-        {{- range $key, $value := .Values.podAnnotations }}
+        {{- range $key, $value := .Values.runner.podAnnotations }}
         {{ $key }}: {{ $value | quote }}
         {{- end }}
       {{- end }}
     spec:
       containers:
       - name: runner
-        image: {{ printf "%s:%s"  .Values.image.repository .Values.image.tag | quote }}
-        imagePullPolicy: {{ .Values.image.pullPolicy | quote }}
+        image: {{ printf "%s:%s"  .Values.runner.image.repository .Values.runner.image.tag | quote }}
+        imagePullPolicy: {{ .Values.runner.image.pullPolicy | quote }}
         env:
-        {{- if eq .Values.orgRunner "true" }}
+        {{- if .Values.runner.orgName }}
         - name: ORG_RUNNER
           value: "true"
         - name: ORG_NAME
-          value: {{ .Values.orgName | quote }}
+          value: {{ .Values.runner.orgName | quote }}
         {{- else }}
         - name: ORG_RUNNER
           value: "false"
         - name: REPO_URL
-          value: {{ .Values.repoUrl | quote }}
+          value: {{ .Values.runner.repoUrl | quote }}
         {{- end }}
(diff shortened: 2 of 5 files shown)


andylightfoot (commented):
LGTM, you just need to update the chart version in Chart.yaml.

charts/github-actions-runner/org_runner.yaml
```diff
@@ -0,0 +1,5 @@
+---
```
xoen, line 1: what are these `org_runner.yaml` and `repo_runner.yaml` files for and where are they used?

damacus: See what what in the PR description :) (testing)

xoen: right, soz 🤦

xoen: It actually makes me wonder if this should go in a README and/or in an `examples/` directory or both. (but now I understand this is just an example of values for when this chart is installed)

charts/github-actions-runner/templates/deployment.yaml
```diff
@@ -19,41 +19,43 @@ spec:
...
           value: "false"
         - name: REPO_URL
-          value: {{ .Values.repoUrl | quote }}
+          value: {{ .Values.runner.repoUrl | quote }}
         {{- end }}
         - name: ACCESS_TOKEN
           valueFrom:
             secretKeyRef:
-              name: {{ .Values.secretAccessTokenName | quote }}
+              name: {{ .Values.runner.secretAccessTokenName | quote }}
```
xoen, line 48: 🤔  this is mentioning this secret but I don't see this in the helm chart, are we planning to create the secret manually?

We have similar helm charts were a credential is passed in a value and this is stored in the secret, all create with Helm which is nice as we don't have to manually write the `Secret` YAML file and do the `kubectl apply` to create the secret which is potentially error-prone.

Maybe I'm missing something here, sorry if I am.

xoen (approved):
Hey @damacus looks good to me 👍 

I left my usual annoying comments/questions.

I guess the main one I have is that the `Secret` containing the GH token is only mentioned but not created with Helm ATM (unless I'm missing something). Ideally we want the helm install to also create the secret as it's reduce the manual things to be done.

charts/github-actions-runner/templates/deployment.yaml
```diff
@@ -19,41 +19,43 @@ spec:
...
         - name: DOCKER_HOST
-          value: tcp://localhost:2376
-        {{- if .Values.runnerNamePrefix }}
+          value: {{ .Values.dockerHost }}
+        - name: RUNNER_GROUP
+          value: {{ .Values.runner.runnerGroup }}
+        {{- if .Values.runner.runnerNamePrefix }}
         - name: RUNNER_NAME_PREFIX
-          value: {{ .Values.runnerNamePrefix | quote }}
+          value: {{ .Values.runner.runnerNamePrefix | quote }}
```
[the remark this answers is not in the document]
damacus: RUNNER_NAME_PREFIX is used unless RUNNER_NAME is set in the values. So that's doing what it should be 👍 

We can probably look at cleaning those helpers up in another PR, for now we're getting the correct names in the UI 

e.g. `github-actions-runner-Gm3hN5UzGxe20`

charts/github-actions-runner/templates/deployment.yaml
damacus: As per our conversation on Slack, the secret is created in the Flux config where we pull the secret from. If we put it in this chart we'd need to encrypt the file with SOPS of git-crypt, neither of which are ideal for a public repository. 

Putting it in the Flux config will allow us to change the secret type in the future to something even more secure like sealed secrets.

PR #487, opened 2020-11-03
Fix CHANGELOG markdown

damacus:


xoen (commented):
We've been using the markdown format of [Keep a Changelog](http://keepachangelog.com/en/1.0.0/), I'm not sure what are the advantages of moving to a slightly different format TBH.

Having said that, if you have strong opinions in favour of this I will not object.

damacus:
Not strong opinions, but the formatting of that document is horrible & the whole Add/Changed/Removed as a header is overkill IMHO if you write good messages. 

Often the sections are like this which seems kinda silly

```markdown
## Added
- Added something
```

PR #488, opened 2020-11-03
Stop quoting the docker host variable

damacus:
Fixes the error:
Status: invalid argument "\"tcp://0.0.0.0:276\"" for "-H, --host" flag: Invalid bind address format: "tcp://0.0.0.0:2376"


andylightfoot (approved):
LGTM

PR #489, opened 2020-11-10
webapp 2.3.11: pull fluent-bit image from AWS ECR

xoen:
It's the same image used before and same tag also but it's hosted on
our AWS ECR repository so that we don't hit the DockerHub rate limits.


damacus (approved):

xoen:
Thanks for the review ❤️

PR #490, opened 2020-11-11
Silently drop all kube-probe requests

tomskelley-gjs:
Set maximum log retries to be 10 - default is 17 (bit random?), but this has caused problems with queued logs overwhelming the Elastic.co ingress nodes

Fixes:

https://trello.com/c/79iOACmg/845-define-reasonable-retry-limit-for-logs-and-apply-it-via-chart
https://trello.com/c/SnaBMHdz/840-turn-off-healthz-reporting-for-k8s-pods


andylightfoot (approved):
LGTM

PR #492, opened 2020-11-13
Image pull secrets

damacus:
- [x] airflow-k8s
- [x] airflow-sqlite
- [x] aws-login
- [x] buckets-archiver
- [x] calicoctl
- [x] cluster-autoscaler
- [x] concourse
- [x] init-user
- [ ] dns-controller
- [ ] kube2iam
- [ ] kuberos
- [x] logstash-auth0
- [x] nfs backup
- [x] pod cleaner
- [x]  prometheus operator (grafana pod)
- [x] RStudio
- [x]  jupyterlab (job bash profile)

The following didn't have any Helm charts to update: 

- [ ] dns-controller
- [ ] kube2iam
- [ ] kuberos

### DIFF
--- a/charts/airflow-k8s/templates/create-nfs-dirs.yml
+++ b/charts/airflow-k8s/templates/create-nfs-dirs.yml
@@ -11,6 +11,12 @@ spec:
     metadata:
       name: {{ .Release.Name }}-create-nfs-dirs
     spec:
+      {{- with .Values.image.pullSecrets }}
+      imagePullSecrets:
+      {{- range . }}
+        - name: {{ . }}
+      {{- end }}
+      {{- end }}
       containers:
       - name: "create-nfs-dirs"
         restart: Never
--- a/charts/airflow-sqlite/Chart.yaml
+++ b/charts/airflow-sqlite/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v1
 description: Airflow with sqlite. Tasks are run as pods
 name: airflow-sqlite
 icon: https://airflow.incubator.apache.org/_images/pin_large.png
-version: 0.4.1
+version: 0.5.0
 appVersion: "Airflow: 1.10.10, Python: 3.7"
--- a/charts/airflow-sqlite/Chart.yaml
+++ b/charts/airflow-sqlite/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v1
 description: Airflow with sqlite. Tasks are run as pods
 name: airflow-sqlite
 icon: https://airflow.incubator.apache.org/_images/pin_large.png
-version: 0.4.0
+version: 0.4.1
 appVersion: "Airflow: 1.10.10, Python: 3.7"
--- a/charts/calicoctl/CHANGELOG.md
+++ b/charts/calicoctl/CHANGELOG.md
@@ -1,0 +1,10 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
+and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+## [0.2.0] - 2020-11-13
+
+- Adds image pull secrets so images can be pulled from Docker Hub with authentication
--- a/charts/cluster-autoscaler/CHANGELOG.md
+++ b/charts/cluster-autoscaler/CHANGELOG.md
@@ -1,27 +1,34 @@
 # Changelog
+
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [0.2.0] - 2020-11-13
+
+- Adds image pull secrets so images can be pulled from Docker Hub with authentication
+
 ## [0.1.5] - 2019-10-21
-### Changed
+
 - Upgraded cluster autoscaler to lockstep with k8s version in new environment from `1.12.8` to `1.13.8`
 
 ## [0.1.4] - 2019-10-21
-### Changed
+
 - Updated clusterrole api-group for replicasets
 - Upgraded cluster autoscaler to lockstep with k8s version in __OLD__ environment from `1.3.5` to `1.12.8`
 
 ## [0.1.3] - 2019-03-13
-### Changed
-- Increased CPU requests and limits from 100m to 200m and 500m respectively  
+
+- Increased CPU requests and limits from 100m to 200m and 500m respectively
 
 ## [0.1.2] - 2018-11-20
+
 ### Added
+
 - Modified image tag from `v1.2.2` to `v1.3.5` after cluster upgrade. See [version table](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler#releases)
 
 ## [0.1.1] - 2018-11-20
-### Added
+
 - Added `Rolebinding` that allows __cluster-autoscaler__ to read the `configmap: cluster-autoscaler-status` in `namespace: default`
 - Added `--namespace=default` flag as __cluster-autoscaler__ assumes that it's deployed to `namespace: kube-system`
--- a/charts/init-user/CHANGELOG.md
+++ b/charts/init-user/CHANGELOG.md
@@ -1,32 +1,44 @@
 # Changelog
+
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [0.2.0] - 2020-11-13
+
+- Adds image pull secrets so images can be pulled from Docker Hub with authentication
+
 ## [0.1.6] - 2019-09-10
+
 ## User Read Rolebinding
+
 - Add a `RoleBinding` in user namespaces for __user-read__ `ClusterRole` and team.
 
-## [0.1.5] - 2019-08-19 
+## [0.1.5] - 2019-08-19
+
 ## User Support role
+
 - Allow members of the user-support github team to have admin access to user namespaces
 
+## [0.1.4] - 2018-03-13
 
-## [0.1.4] - 2018-03-13 
 ## Lock down namespace assume role
+
 - Restrict a users namespace to only assuming the role from their own user
 
+## [0.1.2] - 2018-03-13
 
-## [0.1.2] - 2018-03-13 
-## OIDC Username Claim
+# OIDC Username Claim
+
 - Add idp prefix to username in Clusterrolebinding
 
-
 ## [0.1.1] - 2018-01-09
+
 ## Add Clusterrolebinding
+
 - This assigns the admin user-facing role to a user, scoped to the user's namespace
 
-
 ## [0.1.0] - 2017-07-13
+
 ## Copy charts from ops
--- a/charts/logstash/CHANGELOG.md
+++ b/charts/logstash/CHANGELOG.md
@@ -1,22 +1,24 @@
 # Changelog
+
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [1.1.0] - 2020-11-13
+
+- Adds image pull secrets so images can be pulled from Docker Hub with authentication
 
 ## [1.0.0] - 2018-09-27
-## Changed
+
 Added (optional) TLS block in ingress resource.
 You can set the `ingress.addTlsBlock` value to `false` if your
 ingress-controller doesn't work when ingress resources have this block (e.g. [traefik](https://traefik.io) doesn't seem to work)
 
-
 ## [0.1.1] - 2018-07-09
-## Upgrade Logstash
+
 - Bumping image tag to 6.3.1 (from 5.6.10) to match elasticsearch version
 
-
 ## [0.1.0] - 2018-06-26
-### Initial Commit
+
 - Initial chart development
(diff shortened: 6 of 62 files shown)


damacus:
All charts added via a separate commit for easier diffing. As I expect this PR to get rather large/confusing

charts/airflow-k8s/templates/create-nfs-dirs.yml
```diff
@@ -11,6 +11,12 @@ spec:
     metadata:
       name: {{ .Release.Name }}-create-nfs-dirs
     spec:
+      {{- with .Values.image.pullSecrets }}
```
xoen, line 14: (Same goes for rest of the diff so I'll just comment on first helm chart)

I don't see any default `image.pullSecrets` in the `values.yaml` file, I know these are credentials so there is no meaningful default but it helps understand the structure of the values the helm chart expects. This leads me to the second thought...

I'm not quire sure I follow this `with`/`range .`/`{{ . }}` looks quite clever but maybe not very readable. I'd rather have a more plain template unless there are very compelling reasons to use `with`.

The other observation I have is that AFAICT this template now expects `.Values.image.pullSecrets` to always be passed in. Maybe I'm missing something (I'm not that familiar with `with`, but I had a look at the Helm documentation and it seems to just change the scope nothing else)

It would be better to maintain compatibility with the past and only add `imagePullSecrets` if `.Values.image.pullSecrets` is provided, this also give us more flexibility when deploying as we don't have to wait for the config to update, we can just merge and when the values config are updated we can re-deploy and they'll be added to the `spec`.

(hope this makes sense)

charts/airflow-sqlite/Chart.yaml
```diff
@@ -2,5 +2,5 @@ apiVersion: v1
 description: Airflow with sqlite. Tasks are run as pods
 name: airflow-sqlite
 icon: https://airflow.incubator.apache.org/_images/pin_large.png
-version: 0.4.0
+version: 0.4.1
```
xoen, line 5: Assuming I understand the template diff correctly, isn't the `.image.pullSecrets` value mandatory now? If that's the case it would require a change to the major version to follow semantic versioning.

I still think the best thing to do the `.image.pullSecrets` value should be optional so that it's used only if provided and this way it's a minor bump (not patch, as it adds a feature, the support for `imagePullSecrets`.

damacus: It's now, _sort of_, mandatory. 

If you give the block an empty array then it doesn't print out anything. 

Thanks for spotting the few I didn't add the sensible default in!

charts/rstudio/values.yaml
```diff
@@ -1,8 +1,12 @@
 username: ""
 toolsDomain: "tools.EXAMPLE.COM"
+image:
+  pullSecrets: []
 aws:
   iamRole: ""
   defaultRegion: "eu-west-1"
+image:
+  pullSecrets: []
```
xoen, line 9: Are these duplicated in this `values.yaml` file or is it GH diff broken?

charts/rstudio/CHANGELOG.md
```diff
@@ -1,81 +1,77 @@
 # Changelog
+
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [2.2.7] - 2020-11-13
```
xoen, line 8: As we follow semantic versioning, it should be `2.3.0` as this version adds a feature in a backward compatible way (the ability for the chart to specify the `imagePullSecrets`). Same goes for all the other versions, sorry :(

charts/pod-cleaner/templates/cronjob.yaml
```diff
@@ -17,10 +17,16 @@ spec:
...
+          imagePullSecrets:
+          {{- range . }}
+            - name: {{ . }}
+          {{- end }}
+          {{- end }}
           serviceAccountName: {{ .Release.Name }}
           containers:
           - name: failed-pods-cleaner
-            image: "{{ .Values.image }}"
+            image: "{{ .Values.imageValue }}"
```
xoen, line 29: Why this changed? I'm wary of non-related changes creeping in.

A part from these kind of changes to the values format/structure may have side effects if somehow this chart is installed by CI/CD or similar (don't think it's the case for this helm chart)

But in this case the `values.yaml` structure is different than the value in the template (shouldn't it be `.Values.image.value`?)

Are we introducing too many changes at once?

damacus: We have to introduce this change as the currently image is a string, whereas it needs to be a map. 

Therefore I've moved the image name to `image.tag `and `image.repository` to fall inline with the other repositories.

charts/rstudio/values.yaml
damacus: That's weird. I've just check the diff and yeah. GitHub is actually wrong in this case(!)

charts/logstash/values.yaml
```diff
@@ -6,6 +6,7 @@ image:
   tag: 6.3.1
   pullPolicy: IfNotPresent
   port: 8080
+  pullsecrets: []
```
xoen, line 9: This should be `pullSecrets` (Helm values are case sensitive according to their documentation).

charts/pod-cleaner/templates/cronjob.yaml
xoen: You can break the compatibility if required (as long as we bump the major version of this helm chart release).

My point was about the fact that values file has a `.Values.image.value` value but the template uses `.Values.imageValue` - so unless I'm missing something here this is broken.

xoen (dismissed):
Overall looks good. Putting aside for a moment the questions of introducing all these changes at once and the non-semantic versions it looks good but there are a couple of things that need to change.

Specifically the `image.pullsecrets` value and `imageValue` value which unless I'm missing something are not correct (see comments)

@andylightfoot / @tomskelley-gjs a lof of changes here, could you please cast your RealOps eyes on this in case I missed anything?

charts/logstash/values.yaml
damacus: Oh man, I can't believe I mistyped that! Great spot there

charts/pod-cleaner/templates/cronjob.yaml
damacus: I'll bump a major version here to pull it inline with all of the other images 👍

damacus: OK fixed pod cleaner. Pushed it as `0.2.0` as semVer states that anything before 1.0.0 doesn't have a stable API and can contain breaking changes 👍

xoen: Doesn't this need to be `.Values.image.value` tho? Maybe I'm missing something but [on line 6 and 7 of the `charts/pod-cleaner/values.yaml` file the value is under `image.value` rather than being called `imageValue`. 

Maybe I'm not explaining myself very well here sorry. Or maybe I'm missing something.

charts/calicoctl/CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
+and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+## [0.1.0] - 2020-11-13
```
andylightfoot, line 8: This should be 0.1.1

charts/cluster-autoscaler/CHANGELOG.md
```diff
@@ -1,27 +1,34 @@
 # Changelog
+
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [0.2.0] - 2020-11-13
```
andylightfoot, line 8: or 0.1.6?

charts/init-user/CHANGELOG.md
```diff
@@ -1,32 +1,44 @@
 # Changelog
+
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [0.2.0] - 2020-11-13
```
andylightfoot, line 8: Maybe 0.1.7?

charts/rstudio/values.yaml
```diff
@@ -1,8 +1,12 @@
 username: ""
 toolsDomain: "tools.EXAMPLE.COM"
+image:
+  pullSecrets: []
 aws:
   iamRole: ""
   defaultRegion: "eu-west-1"
+image:
```
andylightfoot, line 8: why does this appear twice?

charts/rstudio/CHANGELOG.md
```diff
@@ -1,81 +1,77 @@
 # Changelog
+
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [2.3.0] - 2020-11-13
```
andylightfoot, line 8: this doesn't match the version in Chart.yaml

charts/logstash/CHANGELOG.md
```diff
@@ -1,22 +1,24 @@
 # Changelog
+
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [1.1.0] - 2020-11-13
```
andylightfoot, line 8: doesn't match the version in chart.yaml

charts/nfs-backup/CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
+and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+## [0.3.0] - 2020-11-13
```
andylightfoot, line 8: doesnt match chart.yaml

andylightfoot (commented):
Can you update all the Changelogs to match the versions in Chart.yaml?

charts/init-user/CHANGELOG.md
damacus: @xoen suggested we minor bump them instead of patch. These did match 😬

xoen: @andylightfoot yes I suggested that.

If the the change to an helm chart is to add a feature in a backward compatible way (e.g. to add the ability to authenticate to a container registry) we should follow semantic version and increase the minor number rather than the patch (that you'd bump for backward compatible bug fixes).

andylightfoot: Yeah my point isn't about that, it is that the version in the Changelog should match the version in Chart.yaml.

charts/rstudio/values.yaml
xoen: @andylightfoot I think it may be a quirk of the GitHub UI diff? I noticed the same here: 🤷 https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/492/files/8480f45b11b6b222054ffca8f5e9349facfd62c8#r524331737 

Actually...I just pulled this branch locally as I was curious and it seems like although [this commit](https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/492/commits/31b562b586f83ae44bf2d505269fcc0cb70bc6ce) adds it once it was also added in this [other commit](https://github.com/ministryofjustice/analytics-platform-helm-charts/commit/93109d98996dd48d4d0c7614a7fae8130d5dc078) - shouldn't cause any problems but it's still duplicated.

charts/rstudio/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
 description: RStudio with Auth0 authentication proxy
 name: rstudio
-version: 2.2.6
+version: 2.3.0
 appVersion: "RStudio: 1.2.1335+conda, R: 3.5.1, Python: 3.7.1, patch: 10"
```
andylightfoot, line 5: Will this mean that the description of both charts is the same in the Control panel drop down?

charts/pod-cleaner/templates/cronjob.yaml
andylightfoot: Yep image: "{{ .Values.imageValue }}" should become image: "{{ .Values.image.value }}"

charts/rstudio/Chart.yaml
xoen: @andylightfoot the description yes. But I think that's fine as:
- this helm chart doesn't change the underlying RStudio Docker image (so `appVersion` shouldn't really change)
- CP also shows the helm chart version so user will see something is different (although it's an internal change and nothing else should have changed)

I still wonder if this PR is updating too many helm charts and some don't really need to change - e.g. as they're currently not using helm docker image in DockerHub 🤷‍♂️

andylightfoot: Cool thanks @xoen 

Yeah I agree we should have done a few smaller PRs targeting the key charts using images from DockerHub.

damacus: Definitely...!

xoen:
I think we can now close this ☺️ - it's superseded by [this other PR (merged)](https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/495) plus [this one for the `jupyter-lab` chart](https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/497) and possibly a couple of others later on.

So closing this. Thanks for your work on this @damacus - the other PR's commits were mostly cherry-picked from this one and/or highly inspired by this work here ☺️ ⭐

xoen:
PS: If there is something else that was in this that I forgot we can always open another PR ☺️  (and also sorry if that's the case).

PR #495, opened 2020-11-23
Support for `imagePullSecrets`

xoen:
The point of this PR is really reduce the number of pull request made to DockerHub and enable this either by adding authentication support (using k8s' `imagePullSecrets`) or by just using the same docker image hosted in our private AWS ECR registry (which simpler and better as it wouldn't require us to add a secret in the namespace as well)

I've done this in alphabetical order so that I don't get too confused about what's done and what's not.
I've got to the charts starting with "P" so far. 

Also, as said in Slack, if we're happy with these changes so far we can just merge this PR and continue this in another PR.

**NOTE**: This is a subset (with alterations!) of Dan's PR here: https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/492


andylightfoot (approved):
LGTM

xoen:
FYI: Merging this PR/batch of helm charts changes and continue in a new PR so that these changes are in `main` now (in case we get distracted by other problems, etc...)

PR #496, opened 2020-11-24
Bump logstash version to 7.9.1

tomskelley-gjs:
This is the same version we're running in the elastic.co cluster. Add some parameters which Elastic have changed for..... reasons..... 

Add some filters to prevent certain things being ingested into Elastic.co


andylightfoot (approved):
I can't really add very much on the fluentd config but everything looks reasonable

PR #498, opened 2020-12-11
Allow jupyter file viewer to see hidden files

isichei:
Think this is everything needed for an update, but let me know if I missed anything.


andylightfoot (approved):
I did a quick test on my Jupyter instance and it seemed to workfine so LGTM.

PR #502, opened 2021-01-07
[rstudio] Increase the initial delay on both the livenessProbe and

damacus:
readinessProbe

### DIFF
--- a/charts/rstudio/CHANGELOG.md
+++ b/charts/rstudio/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [3.0.2] - 2021-01-07
+
+- Increase the initial delay on both the livenessProbe and readinessProbe to stop the container restarting before it's has a chance to initialize
+
 ## [3.0.1] - 2021-01-07
 
 - Adds the secure-cookie-key to the server so that auth-proxy can talk to RStudio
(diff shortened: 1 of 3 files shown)


damacus:
It's seems so on the platform. I'm giving it a wide birth for now. It actually starts up almost instantly locally. But I assume there's some chown'ing going on in the background that is causing it.

andylightfoot:
> It's seems so on the platform. I'm giving it a wide birth for now. It actually starts up almost instantly locally. But I assume there's some chown'ing going on in the background that is causing it.

OK I think we need to investigate what is causing this as it isn't really ideal to increase this timing as it means that every time R Studio starts we are adding additional time before the probes actually start checking if it is ready and then also additional time for the probe to mark it as ready. 

Maybe edit the settings in the deployment for now rather than changing the values in the helm chart?

ntoll (approved):
LGTM

andylightfoot:
> > It's seems so on the platform. I'm giving it a wide birth for now. It actually starts up almost instantly locally. But I assume there's some chown'ing going on in the background that is causing it.
> 
> OK I think we need to investigate what is causing this as it isn't really ideal to increase this timing as it means that every time R Studio starts we are adding additional time before the probes actually start checking if it is ready and then also additional time for the probe to mark it as ready.
> 
> Maybe edit the settings in the deployment for now rather than changing the values in the helm chart?

erm why has this been merged?

damacus:
I was debugging this with @ladyrassilon & @ntoll and the low timeout of 5 seconds made kubernetes restart the container multiple times before the probe got lucky and succeeded. 

30 seconds seems like a reasonably low initial delay and that's worked in testing with @ntoll and my deployments 👍🏼

PR #505, opened 2021-01-14
[rstudio] Change the default USER to rstudio

damacus:
## JIRA Tickets

ANPL-239


davidread (approved):
I like this simplification. Users will cope, I think.
When we deploy we may find issues, but let's try it with the gradual roll-out planned.

PR #506, opened 2021-01-25
Revert the auth-proxy USERNAME change

damacus:


ntoll (approved):
LGTM --- ahhh.... that's what `values.username` is used for...! ;-)

PR #512, opened 2021-02-04
[ANPL-239] Change the chown home command to a config map

damacus:
We mount the config map as a script into the container and run it as
part of the container


andylightfoot (approved):
Looks fine to me

damacus:
🤞🏼

PR #515, opened 2021-03-29
[ANPL-365] Use the new NGINX Proxy container

damacus:


ntoll (approved):
nginx the things :shipit:

PR #518, opened 2021-03-31
Ensure resolver envar is set.

ntoll:


charts/rstudio/values.yaml
```diff
@@ -1,18 +1,19 @@
-username: "test"
+username: "test"  # TODO: make this the right name.
```
damacus, line 1: This is only in values.yaml so tests pass. IIRC, It gets passed into the template from the Control Panel

PR #520, opened 2021-04-20
WiP - surely I just needed to update the versions..?

ntoll:
So this doesn't work. I'm not sure why, although I have some ideas.

1) This helm chart is based on the continuing work that Dan did. I'm not sure if his changes to the helm chart will work with the new work on the docker images I did.
2) I'm just running the wrong sort of commands when trying to test `helm`... 

```
$ helm upgrade --install --wait --force rstudio-ntoll rstudio --version 3.1.3 --namespace user-ntoll --values rstudio/values.yaml --set proxy.auth0.domain=alpha-analytics-moj.eu.auth0.com --set proxy.auth0.clientId=$AUTH0_CLIENT_ID --set proxy.auth0.clientSecret=$AUTH0_CLIENT_SECRET
Release "rstudio-ntoll" does not exist. Installing it now.
Error: release rstudio-ntoll failed: timed out waiting for the condition
```

3) Nicholas does devops is like Godzilla does needlework. ;-)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .DS_STORE
 .vscode/
 .idea/
+.envrc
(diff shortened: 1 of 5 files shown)


ntoll:
Worth pointing out that the example command is run within the `analytics-platform-helm-charts` repository, so I'm assuming it's running my * local* version of the chart.

tomskelley-gjs (approved):
LGTM.

PR #522, opened 2021-05-04
ANPL-400 Add option to disable secret creation in the control panel helm chart

andylightfoot:
When we move to EKS we would like to avoid having to create Kubernetes secrets with Helm charts, as this requires us to provide secret values, in the clear as inputs to helm charts.

Instead we would like to have the option of creating the Kubernetes secrets from values stored in AWS secrets manager.

This change allows us to disable to create of Kubernetes secrets for in the control panel helm chart by setting a secrets.create to false.

By default all Kubernetes secrets will continue to be created as in previous versions of the chart.


neillturner (approved):
seems an odd way to do secrets by passing the values to the helm chart as they are visible in the setup of the call to the helm chart. but i guess that's how they did it. -)

PR #527, opened 2021-06-09
ANPL-479 - Update RStudio helm chart to use nginx proxy image with configurable redirect url

andylightfoot:
Update RStudio helm chart to use the nginx proxy image with a configurable redirect url


tomskelley-gjs (approved):
LGTM

PR #528, opened 2021-06-10
ANPL-496 - do not merge

tomskelley-gjs:
--
Rename files called efs-persistent...... to nfs-persistent..... as they referred to softnas NFS
Add files called efs-persistent..... which do pertain to EFS
Default to dev EFS volume so no changes are needed to cpanel in order to get this working

Provision User
--
Rename pvc to match init user pvc so the same rstudio chart can be used for both


neillturner (approved):

andylightfoot (approved):
👍

PR #536, opened 2021-06-18
ANPL-480 - Add username to subPath of EFS mount in Jupyterlab chart

andylightfoot:
Add username to subPath of EFS mount in Jupyterlab chart


tomskelley-gjs (approved):
LGTM

PR #538, opened 2021-06-22
ANPL-498 Add Renviron configmap to Rstudio chart

andylightfoot:
In order for RStudio users to be able to access AWS resources from RStudio, they need to be able to get AWS credentials by assuming their IAM role.

- For users running on EKS we will grant AWS access using IAM for Service Accounts.

- For users running on the legacy (kops) clusters, we will continue to use use kube2iam to grant AWS access.

RStudio sessions run under a different user from the RStudio container (which runs as root).

This means that for users running on EKS we need to set the  AWS_ROLE_ARN and AWS_WEB_IDENTITY_TOKEN_FILE environment variables for the R Studio session.

This can be done in a few different way, but the more flexible will be to mount a .Renviron file in the root of a users home directory . This will will contain the AWS_ROLE_ARN and AWS_WEB_IDENTITY_TOKEN_FILE environment variables for the R Studio session.

For more details see https://rviews.rstudio.com/2017/04/19/r-for-enterprise-understanding-r-s-startup/

By default we will not mount a .Renviron file, which will ensure this is not mounted for RStudio users on the legacy (kops) clusters.

### DIFF
--- a/charts/rstudio/CHANGELOG.md
+++ b/charts/rstudio/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 # [4.1.2] - 2021-06-21
-- Mount .Renviron file to configure AWS credentials environment variables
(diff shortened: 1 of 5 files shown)


tomskelley-gjs (approved):
LGTM

PR #540, opened 2021-06-25
Turn off prompt for conda from reticulate so that it's NEVER prompted for EVER EVER AGAIN \o/

tomskelley-gjs:


andylightfoot (approved):
LGTM

PR #541, opened 2021-06-25
ANPL-409  Move helm chart automation to GitHub actions

neillturner:
add chart-releaser


andylightfoot (approved):
It looks pretty simple so we should test it and run in parallel with the current workflow.

PR #542, opened 2021-07-28
ANPL-541 Deploy all charts in analytics-platform-helm-charts repo to

neillturner:
change to trigger release of charts.  so i can trigger a deploy of all charts to gh-pages. I tested this on my own repo https://github.com/neillturner/test-github-script so it should create a release for each chart and a reference to it in the gh-pages index.html that will get created. 

in .github/workflows/charts-lint-test-release.yaml  in the on clause path - .github/workflows/charts-lint-test-release.yaml forces it to run the release part
i tested it and it will cause all the charts to create releases


andylightfoot (approved):
YOLO

PR #560, opened 2021-09-09
ANPL-651 - Add user rolebinding to the provision user chart

andylightfoot:
Add user rolebinding to the provision user chart

This allows a user to access their namespace. It is used by the Control panel when they log in.


tomskelley-gjs (approved):
LGTM

